PR #55, opened 2018-06-20
Start session dialog

corbinlc:
give users better feedback of how start session is progressing
also get rid of collecting long result string that we don't really use

### DIFF
--- a/app/src/main/java/tech/ula/ui/SessionListActivity.kt
+++ b/app/src/main/java/tech/ula/ui/SessionListActivity.kt
@@ -18,6 +18,7 @@ import android.preference.PreferenceManager
 import android.support.v4.app.ActivityCompat
 import android.support.v4.content.ContextCompat
 import android.support.v7.app.AppCompatActivity
+import android.util.Log
 import android.view.ContextMenu
 import android.view.Menu
 import android.view.MenuItem
@@ -83,6 +84,11 @@ class SessionListActivity : AppCompatActivity() {
             if (downloadedId != null)
                 downloadedList.add(downloadedId)
         }
+    }
+
+    private val FILESYSTEM_EXTRACT_LOGGER = { line: String -> Int
+        this@SessionListActivity.runOnUiThread({text_session_list_progress_update_line_2.text = getString(R.string.progress_setting_up_extract_text,line)})
+        0
     }
 
     private val fileManager by lazy {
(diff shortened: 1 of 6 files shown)


app/src/main/java/tech/ula/ui/SessionListActivity.kt
```diff
@@ -85,6 +85,11 @@ class SessionListActivity : AppCompatActivity() {
         }
     }
 
+    private val FILESYSTEM_EXTRACT_LOGGER = { line: String -> Int
+        this@SessionListActivity.runOnUiThread({text_session_list_progress_update_line_2.text = getString(R.string.progress_setting_up_extract_text,line)})
```
MatthewTighe: Did you try this without `@SessionListActivity`? I'm pretty sure `this` should refer to SessionListActivity anyway due to lazy evaluation.

app/src/main/res/layout/activity_session_list.xml
```diff
@@ -57,8 +57,22 @@
...
             android:layout_height="wrap_content"
-            android:layout_margin="8dp"
+            android:layout_margin="4dp"
             app:layout_constraintTop_toBottomOf="@id/progress_bar_session_list"
+            app:layout_constraintBottom_toTopOf="@+id/text_session_list_progress_update_line_2"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintEnd_toEndOf="parent" />
+
+        <TextView
+            android:id="@+id/text_session_list_progress_update_line_2"
```
MatthewTighe: I think we should use a more descriptive name for this. Maybe like `session_list_progress_step` and `session_list_progress_details`?

app/src/main/java/tech/ula/ui/SessionListActivity.kt
corbinlc: I did not.  I read through and example and it is just something they did and we do it in other places like ...
val downloadManager = DownloadUtility(this@SessionListActivity, archType, distType)

app/src/main/res/layout/activity_session_list.xml
corbinlc: good point

app/src/main/java/tech/ula/ui/SessionListActivity.kt
corbinlc: Here:
while (downloadList.size != downloadedList.size) {
                    
  this@SessionListActivity.runOnUiThread({text_session_list_progress_update_line_2.text = getString(R.string.progress_downloading_out_of,downloadedList.size,downloadList.size)})
                    delay(500)
                }

I have to do it because this is a CorroutineScope

corbinlc: Ah, that is why the other line of code i mentioned needs it too.
ok, this one place that you mentioned, should not need it

corbinlc:
Please look at this again.  Thanks!

PR #72, opened 2018-06-29
Navigation component and fragmentation.

MatthewTighe:


app/src/main/java/tech/ula/ui/SessionListFragment.kt
```diff
@@ -135,173 +156,107 @@ class SessionListActivity : AppCompatActivity() {
...
-        val intent = Intent(this, SettingsActivity::class.java)
-        startActivity(intent)
+    private fun stopService(session: Session): Boolean {
+        if(session.active) {
+            session.active = false
+            sessionListViewModel.updateSession(session)
+            val view = list_sessions.getChildAt(sessionList.indexOf(session))
+            view.image_list_item_active.setImageResource(R.drawable.ic_block_red_24dp)
+
+            serverUtility.stopService(session)
```
corbinlc: I believe this call of stopService is redundant with what is done inside ServerService when the type is kill

PR #90, opened 2018-07-10
Ubuntu addition

corbinlc:
There is definitely some clean up needed here, but this is our first push towards auto updating assets and taking them from github.  

Also, adds back in ubuntu as a choice as it will need the auto updating piece to roll out to existing users.

### DIFF
--- a/app/src/main/java/tech/ula/utils/DownloadUtility.kt
+++ b/app/src/main/java/tech/ula/utils/DownloadUtility.kt
@@ -2,16 +2,13 @@ package tech.ula.utils
 
 import android.app.DownloadManager
 import android.content.Context
-import android.content.SharedPreferences
 import android.net.ConnectivityManager
 import android.net.NetworkInfo
 import android.net.Uri
 import android.os.Environment
-import android.util.Log
 import java.io.File
 import java.net.HttpURLConnection
 import java.net.URL
-import java.util.*
 
 class DownloadUtility(val context: Context, val archType: String, distType: String) {
 
@@ -80,6 +77,7 @@ class DownloadUtility(val context: Context, val archType: String, distType: Stri
 
         //only download if this is a newFilesystem or it has been more than a day since we last checked
         val now = java.util.Date().time
+        //TODO: should be store last updated in the session itself
         val lastUpdated = sharedPref.getLong("lastUpdated", 0)
         if (now < (lastUpdated + 86400000L) && !newFilesystem) {
             return false
--- a/app/src/main/java/tech/ula/utils/DownloadUtility.kt
+++ b/app/src/main/java/tech/ula/utils/DownloadUtility.kt
@@ -9,12 +9,13 @@ import android.os.Environment
 import java.io.File
 import java.net.HttpURLConnection
 import java.net.URL
+import java.util.concurrent.TimeUnit
+import java.util.Date
 
 class DownloadUtility(val context: Context, val archType: String, distType: String) {
 
-    companion object {
-        val branch = "master"
-    }
+    private val branch = "master"
+    private val failedConnection = 0L
 
     private val downloadManager: DownloadManager by lazy {
         context.getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager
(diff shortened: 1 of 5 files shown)


app/src/main/java/tech/ula/utils/DownloadUtility.kt
```diff
@@ -57,18 +64,53 @@ class DownloadUtility(val context: Context, val archType: String, distType: Stri
...
+        //only download the rootfs the first time, it cannot simply be updated like other files
+        if(type.contains("rootfs") && !newFilesystem) {
+            return false
+        }
+
+        //only download if this is a newFilesystem or it has been more than a day since we last checked
+        val now = java.util.Date().time
+        //TODO: should be store last updated in the session itself
+        val lastUpdated = sharedPref.getLong("lastUpdated", 0)
+        if (now < (lastUpdated + 86400000L) && !newFilesystem) {
```
MatthewTighe: Can we use something besides a magic number here?

MatthewTighe: This is nitpicky, but I also feel like this logic would be easier to reason about if it were reversed; you even do so in your comment. Something like `if (now > (lastUpdated + Date.ADay()) && newFilesystem` ...

app/src/main/java/tech/ula/utils/DownloadUtility.kt
```diff
@@ -57,18 +64,53 @@ class DownloadUtility(val context: Context, val archType: String, distType: Stri
...
+            return false
+        } else {
+            with (sharedPref.edit()) {
+                putLong("lastUpdated", now)
```
MatthewTighe: If the download fails I don't think we'll check for a newer version if they retry it.

app/src/main/java/tech/ula/utils/DownloadUtility.kt
```diff
@@ -57,18 +64,53 @@ class DownloadUtility(val context: Context, val archType: String, distType: Stri
...
+        }
+
         val (subdirectory, filename) = type.split(":")
         val asset = File("${context.filesDir.path}/$subdirectory/$filename")
-        // TODO more sophisticated version checking
+
+        //update a file if a connection can be made and if the datestamp is different
+        val currentDateStamp = sharedPref.getLong(type, 0)
+        val newDateStamp = urlDateModified(endpoint)
+        if ((newDateStamp != 0L) && (currentDateStamp != newDateStamp)) {
```
MatthewTighe: What does `newDateStamp != 0L` check for?

app/src/main/java/tech/ula/utils/DownloadUtility.kt
```diff
@@ -57,18 +64,53 @@ class DownloadUtility(val context: Context, val archType: String, distType: Stri
...
+            with (sharedPref.edit()) {
+                putLong(type, newDateStamp)
+                commit()
```
MatthewTighe: Again, it seems odd that we're updating timestamps of when we've downloaded things before actually downloading them.

app/src/main/java/tech/ula/utils/DownloadUtility.kt
```diff
@@ -7,36 +7,43 @@ import android.net.NetworkInfo
...
 import android.os.Environment
 import java.io.File
+import java.net.HttpURLConnection
+import java.net.URL
 
 class DownloadUtility(val context: Context, val archType: String, distType: String) {
 
+    companion object {
+        val branch = "master"
+    }
```
MatthewTighe: I think this can just be a private val. I'm not seeing it used in any other files.

app/src/main/java/tech/ula/utils/DownloadUtility.kt
corbinlc: this one is ok because we delete the file, which will cause it to have to update as the end logic is just checking for existence.

app/src/main/java/tech/ula/utils/DownloadUtility.kt
corbinlc: If it returns 0L it means it failed to make connection with the server when checking the timestamp (or the path is busted).  in those scenarios we should just stick with the version we already have.

app/src/main/java/tech/ula/utils/DownloadUtility.kt
corbinlc: that is true.  really need to update this later and just do the early exit here.

app/src/main/java/tech/ula/utils/DownloadUtility.kt
corbinlc: i will create a well named private constant with this value... it is the number of ms in a day... you didn't know that? ;)

app/src/main/java/tech/ula/utils/DownloadUtility.kt
corbinlc: true

app/src/main/java/tech/ula/utils/DownloadUtility.kt
corbinlc: the goal is to exit early and not nest code.

PR #167, opened 2018-08-10
do something on execLocal exceptions

corbinlc:
**Describe the pull request**

Need to handle exceptions related to `wrapWithBusyboxAndExecute`
this is just an initial bid, please review and improve

**Link to relevant issues**

### DIFF
--- a/app/src/main/java/tech/ula/utils/ServerUtility.kt
+++ b/app/src/main/java/tech/ula/utils/ServerUtility.kt
@@ -47,26 +47,42 @@ class ServerUtility(private val execUtility: ExecUtility, private val fileUtilit
     }
 
     private fun startSSHServer(session: Session): Long {
+        var pid: Long
         val targetDirectoryName = session.filesystemId.toString()
         deletePidFile(session)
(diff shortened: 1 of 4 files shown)


app/src/main/java/tech/ula/utils/ServerUtility.kt
```diff
@@ -47,26 +47,42 @@ class ServerUtility(private val execUtility: ExecUtility, private val fileUtilit
...
 
     fun stopService(session: Session) {
         val targetDirectoryName = session.filesystemId.toString()
 
         val command = "../support/killProcTree.sh ${session.pid} ${session.pid()}"
-        execUtility.wrapWithBusyboxAndExecute(targetDirectoryName, command)
+        try {
+            execUtility.wrapWithBusyboxAndExecute(targetDirectoryName, command)
+        } catch (err: Exception) {
+            return
```
luongthomas: Should we return -1 here too?

Also would it be helpful if we logged the exception?

MatthewTighe:
This work is being recreated separately to incorporate recent changes to staging, so I'm closing this PR.

Issue #189, opened 2018-08-22
Do not download rootfs files if they aren't immediately going to be used for extraction.


MatthewTighe:
Currently, rootfs files will be downloaded if they are out of date regardless of whether they are going to be used. Update session control flow to only download them if they are about to be extracted.

PR #190, opened 2018-08-22
Do not download rootfs files unless necessary.

MatthewTighe:
**Describe the pull request**

If local rootfs files were out of date, they would be downloaded even if there was no intention to actually use them, leading to large unnecessary network overheads.

**Link to relevant issues**
#189

### DIFF
--- a/app/src/main/java/tech/ula/utils/SessionController.kt
+++ b/app/src/main/java/tech/ula/utils/SessionController.kt
@@ -8,8 +8,10 @@ import tech.ula.model.entities.Session
 import tech.ula.model.repositories.AssetRepository
 
 class SessionController(
+    private val filesystem: Filesystem,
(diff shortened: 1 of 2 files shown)


app/src/main/java/tech/ula/utils/SessionController.kt
```diff
@@ -35,23 +37,28 @@ class SessionController(
...
+                    if (filesystemUtility.hasFilesystemBeenSuccessfullyExtracted("${filesystem.id}"))
+                        return@filter false
+
+                    if (!forceDownloads && !networkUtility.wifiIsEnabled()) {
+                        wifiRequired = true
+                        dialogBroadcaster("wifiRequired")
+                        return@map listOf<Asset>()
+                    }
                 }
                 needsUpdate
```
luongthomas: Is there a purpose for this boolean variable being here?

Issue #203, opened 2018-08-24
Handle ExecLocal Exceptions


luongthomas:
Properly handle exceptions arising from the function `execLocal`.  This includes logging the exception and relevant information.

PR #204, opened 2018-08-24
Handle exceptions from execLocal errors

luongthomas:
**Describe the pull request**

Bubbling up the error message and logging them.

throw_error_in_execlocal
Example of when exception is thrown in the execLocal function.  This is the error message in logcat.

**Link to relevant issues**
#203

### DIFF
--- a/app/src/main/java/tech/ula/utils/LogUtility.kt
+++ b/app/src/main/java/tech/ula/utils/LogUtility.kt
@@ -19,4 +19,10 @@ class LogUtility {
     fun d(tag: String, message: String) {
         Log.d(tag, message)
     }
+
+    fun runtimeErrorForCommand(functionName: String, command: String, err: Exception) {
+        val errorMessage = "Error while executing " +
+                "`$functionName()` with command: $command \n\tError = $err"
+        this.e(tag = "RuntimeError", message = errorMessage)
+    }
 }
--- a/app/src/main/java/tech/ula/utils/ServerUtility.kt
+++ b/app/src/main/java/tech/ula/utils/ServerUtility.kt
@@ -3,7 +3,11 @@ package tech.ula.utils
 import tech.ula.model.entities.Session
 import java.io.File
 
-class ServerUtility(private val applicationFilesDirPath: String, private val execUtility: ExecUtility) {
+class ServerUtility(
+    private val applicationFilesDirPath: String,
+    private val execUtility: ExecUtility,
+    private val logger: LogUtility = LogUtility()
+) {
 
     fun Process.pid(): Long {
         return this.toString().substringAfter("pid=").substringBefore(",").substringBefore("]").trim().toLong()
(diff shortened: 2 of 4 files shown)


app/src/main/java/tech/ula/utils/ServerUtility.kt
```diff
@@ -47,37 +51,53 @@ class ServerUtility(private val applicationFilesDirPath: String, private val exe
     }
 
     private fun startSSHServer(session: Session): Long {
+        var processPid = -1L
         val targetDirectoryName = session.filesystemId.toString()
         deletePidFile(session)
         val command = "../support/execInProot.sh /bin/bash -c /support/startSSHServer.sh"
-        val process = execUtility.wrapWithBusyboxAndExecute(targetDirectoryName, command, doWait = false)
-        return process.pid()
+        try {
```
MatthewTighe: these could be 
```
return try { ... } 
catch { 
    logger.runtimeErrorForCommand() 
    -1
}
```

app/src/main/java/tech/ula/utils/LogUtility.kt
```diff
@@ -19,4 +19,10 @@ class LogUtility {
     fun d(tag: String, message: String) {
         Log.d(tag, message)
     }
+
+    fun runtimeErrorForCommand(functionName: String, command: String, err: Exception) {
+        val errorMessage = "Error while executing " +
+                "`$functionName()` with command: $command \n\tError = $err"
+        this.e(tag = "RuntimeError", message = errorMessage)
```
MatthewTighe: I don't think you need the explicit reference to `this` in this case, but I don't care if you leave it in.

app/src/main/java/tech/ula/utils/LogUtility.kt
```diff
@@ -19,4 +19,10 @@ class LogUtility {
     fun d(tag: String, message: String) {
         Log.d(tag, message)
     }
+
+    fun runtimeErrorForCommand(functionName: String, command: String, err: Exception) {
```
MatthewTighe: I still prefer a more action-oriented naming for convention type personally. Something like `fun logRuntimeErrorForCommand()`.

app/src/main/java/tech/ula/utils/ServerUtility.kt
luongthomas: That's pretty convenient.  I'll change it to use `return try`

app/src/main/java/tech/ula/utils/LogUtility.kt
luongthomas: I just noticed the codebase has function names that all start with a verb.  I'll stick with that theme.

app/src/main/java/tech/ula/utils/ServerUtility.kt
MatthewTighe: Gets rid of the `var` handily enough as well.

PR #266, opened 2018-09-21
Apps non default user login

luongthomas:
**Describe the pull request**

When filesystem is being created, the user can enter in username, password and vncpassword

**Link to relevant issues**


app/src/androidTest/java/tech/ula/model/MigrationTest.kt
```diff
@@ -94,6 +94,8 @@ class MigrationTest {
         assertEquals(session2, returnedSession2)
         assertFalse(returnedSession1.isAppsSession)
         assertTrue(returnedSession2.isAppsSession)
+        assert(fs.defaultVncPassword == "userland")
```
MatthewTighe: `assert` is actually disabled on the JVM by default. Tests have to be run with special flags in order to get them to fire and I can't remember if that's setup right now. For safeties sake, use stuff like `assertEquals` from Junit.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -83,30 +81,98 @@ class FilesystemEditFragment : Fragment() {
...
+            override fun onTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {}
+        })
+    }
+
     private fun insertFilesystem(): Boolean {
         val navController = NavHostFragment.findNavController(this)
-        if (filesystem.name == "") input_filesystem_name.error = getString(R.string.error_filesystem_name)
-        if (filesystem.name == "") {
-            Toast.makeText(activityContext, R.string.error_empty_field, Toast.LENGTH_LONG).show()
+        if (!isFilesystemParametersCorrect()) {
```
MatthewTighe: I've done this myself tons of places and still catch myself doing it now, but I think this reads better as something like:

`if (!filesystemParametersAreCorrect())`

To me that just matches more closely how I would actually want to say that conditional out loud to someone, and it makes it easier to think about. Personal preference thing though.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -83,30 +81,98 @@ class FilesystemEditFragment : Fragment() {
         }
     }
 
+    fun setupTextInputs() {
+        input_filesystem_name.setText(filesystem.name)
```
MatthewTighe: Does `setText` need to be invoked here, or is it possible to Kotlin property access syntax?

app/src/main/res/layout/frag_filesystem_edit.xml
```diff
@@ -22,13 +22,63 @@
...
+        android:layout_height="wrap_content"
+        android:padding="8dp"
+        app:layout_constraintTop_toBottomOf="@+id/text_input_layout">
+
+        <android.support.design.widget.TextInputEditText
+            android:id="@+id/input_filesystem_username"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:inputType="text"
+            android:text="@string/default_username"
```
MatthewTighe: Let's set these dynamically based on the filesystem, rather than a string resource. That way we only have to change it in one place if we decide to.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
luongthomas: I tried to do `input_filesystem_name.text = filesystem.name` here but it seems that it expects an `Editable` rather than a string.

There are users that recommend using `setText()` to change the text.
https://stackoverflow.com/questions/44493908/setting-text-in-edittext-kotlin/44493995#44493995

app/src/main/res/layout/frag_filesystem_edit.xml
luongthomas: Could you go into a bit more detail about this?

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
MatthewTighe: Yeah, you're totally right. `TextViews` can be set the easier way but not editable ones.

app/src/main/res/layout/frag_filesystem_edit.xml
MatthewTighe: It looks like you're actually already doing that in `setupTextInputs()`, sorry. But that means you should be able to remove the static xml attributes and therefore the string resources as well I believe.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+
+class ValidationUtility() {
+
+    fun isUsernameValid(password: String): Boolean {
+        val pattern: Pattern
+        val matcher: Matcher
+
+        val PASSWORD_PATTERN = "^[a-zA-Z0-9]*\$"
```
MatthewTighe: This name seems odd to me.

luongthomas: Oops it should be renamed to mention `username`

IS is USERNAME_REGEX a better name?

MatthewTighe: `USERNAME_PATTERN` makes the most sense to me, to match the `PASSWORD_PATTERN` below it.

PR #267, opened 2018-09-21
Apps pref dialog

luongthomas:
**Describe the pull request**

When selecting an application, there will be a dialog message to show whether to use SSH or VNC

**Link to relevant issues**


app/src/main/java/tech/ula/ui/AppListFragment.kt
```diff
@@ -158,4 +174,26 @@ class AppListFragment : Fragment() {
     private fun passDataToActivity(data: String) {
         dataPasser.onFragmentDataPassed(data)
     }
-}
+
+    private fun select_client_preference(selectedApp: App) {
```
MatthewTighe: Stick with camel case for kotlin side code. Snake case is used to denote xml entities.

app/src/main/java/tech/ula/ui/AppListFragment.kt
```diff
@@ -104,11 +114,17 @@ class AppListFragment : Fragment() {
                     return@OnItemClickListener
                 }
 
-                val serviceIntent = Intent(activityContext, ServerService::class.java)
-                        .putExtra("type", "startApp")
-                        .putExtra("app", selectedApp)
-                        .putExtra("serviceType", "ssh") // TODO update this dynamically based on user preference
-                activityContext.startService(serviceIntent)
+                val preferredClient = appListViewModel.getAppServiceTypePreference(appList[position])
+                if (preferredClient.isEmpty()) {
```
MatthewTighe: Nice use of the null object pattern. I didn't actually know you could call `isEmpty()` on strings like that.

app/src/main/java/tech/ula/ui/AppListFragment.kt
```diff
@@ -158,4 +174,26 @@ class AppListFragment : Fragment() {
...
+    private fun select_client_preference(selectedApp: App) {
+        lateinit var dialog: AlertDialog
+
+        val clients = arrayOf("SSH", "VNC")
+        var preferredServiceType = "SSH"
+
+        val builder = AlertDialog.Builder(activityContext)
+        builder.setTitle("Always open with:")
+        builder.setSingleChoiceItems(clients, 0) { _, selected ->
+            val choice = clients[selected]
```
MatthewTighe: This could probably be one line and not lose context.

app/src/main/java/tech/ula/model/repositories/AppsRepository.kt
```diff
@@ -6,16 +6,29 @@ import android.util.Log
...
+    private val appsPreferences: AppsPreferences
+) {
     private val refreshStatus = MutableLiveData<RefreshStatus>()
 
     fun getAllApps(): LiveData<List<App>> {
         return appsDao.getAllApps()
     }
 
+    fun getAppServiceTypePreference(app: App): String {
+        return appsPreferences.getAppServiceTypePreference(app.name)
```
MatthewTighe: Seeing this makes me think this preference should be attached to the app entity rather than used as a Android preference. Should talk about this more on Monday.

app/src/main/java/tech/ula/ui/AppListFragment.kt
```diff
@@ -86,6 +96,18 @@ class AppListFragment : Fragment() {
                 is AppItem -> {
                     val selectedApp = selectedItem.app
                     doAppItemClicked(selectedApp)
+
+                    val preferredServiceType = appListViewModel.getAppServiceTypePreference(appList[position])
```
MatthewTighe: appList[position] is broken now. There are more items than apps in the list because of the separators. Use `selectedApp`.

app/src/main/java/tech/ula/ui/AppListFragment.kt
```diff
@@ -179,4 +201,25 @@ class AppListFragment : Fragment() {
...
+
+        val builder = AlertDialog.Builder(activityContext)
+                .setTitle("Always open with:")
+                .setSingleChoiceItems(serviceTypes, 0) { _, selected ->
+                    preferredServiceType = serviceTypes[selected]
+                }
+
+        builder.setPositiveButton("Continue") { _, _ ->
+            appListViewModel.setAppServiceTypePreference(selectedApp, preferredServiceType)
+            Toast.makeText(activityContext, "$preferredServiceType selected", Toast.LENGTH_SHORT).show()
```
MatthewTighe: User facing text needs to be defined in `strings.xml`, but this could also just be removed imo.

app/src/main/java/tech/ula/ui/AppDetailsFragment.kt
```diff
@@ -8,26 +10,49 @@ import android.view.ViewGroup
...
         apps_icon.setImageURI(localFileLocator.findIconUri(app.name))
         apps_title.text = app.name
         apps_description.text = (localFileLocator.findAppDescription(app.name))
+
+        setupPreferredServiceTypeRadioGroup()
+    }
+
+    fun setupPreferredServiceTypeRadioGroup() {
+        val appServiceTypePreference = appsPreferences.getAppServiceTypePreference(app.name)
+        if (appServiceTypePreference == "SSH") {
```
MatthewTighe: I think these strings should be defined as constants on the `AppPreferences` class so we can change these checks to avoid string literals.

PR #302, opened 2018-10-03
Remove default values in filesystem creation page

luongthomas:
**Describe the pull request**

In filesystem edit page, default values for username, password and vncpassword are now blank.  It will be required to fill all values.  

Also the fragment is encapsulated in a scroll view to accommodate screens that are small or are in landscape mode by allowing the text field to be visible to the user when a text field is focused.  It was hidden by the keyboard before.

Also fixed bug where filesystem was able to be created with a blank filesystem name.

**Link to relevant issues**


app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -155,14 +155,17 @@ class FilesystemEditFragment : Fragment() {
...
+        val allCredentialsAreValid = true
         val validator = ValidationUtility()
-        var allCredentialsAreValid = false
         val username = filesystem.defaultUsername
         val password = filesystem.defaultPassword
         val vncPassword = filesystem.defaultVncPassword
 
+        if (filesystem.name.isEmpty()) {
+            input_filesystem_name.error = getString(R.string.error_filesystem_name)
+            return !allCredentialsAreValid
```
MatthewTighe: I'd just return values from this function, rather than a variable. That way future readers don't have to spend any time tracking the logic.

luongthomas: I removed the variable `allCredentialsAreValid` and used just a raw boolean.  I agree it'll make it easier to read

PR #320, opened 2018-10-10
Change name of app startup script.

MatthewTighe:
By renaming the app startup script to a bunch of z's while it is briefly in profile.d (it is deleted immediately on execution), it should execute as the last step no matter what else is in the directory.

### DIFF
--- a/app/src/main/java/tech/ula/utils/FilesystemUtility.kt
+++ b/app/src/main/java/tech/ula/utils/FilesystemUtility.kt
@@ -96,9 +96,11 @@ class FilesystemUtility(
 
     fun moveAppScriptToRequiredLocations(appName: String, appFilesystem: Filesystem) {
         // TODO add error cases
+        val fileNameToForceAppScriptToExecuteLast = "zzzzzzzzzzzzzzzz.sh"
         val appScriptSource = File("$applicationFilesDirPath/apps/$appName/$appName.sh")
         val appScriptSupportTarget = File("$applicationFilesDirPath/${appFilesystem.id}/support/$appName.sh")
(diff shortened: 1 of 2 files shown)


app/src/main/java/tech/ula/utils/FilesystemUtility.kt
```diff
@@ -96,9 +96,11 @@ class FilesystemUtility(
 
     fun moveAppScriptToRequiredLocations(appName: String, appFilesystem: Filesystem) {
         // TODO add error cases
+        val fileNameToForceAppScriptToExecuteLast = "zzzzzzzzzzzzzzzz.sh"
         val appScriptSource = File("$applicationFilesDirPath/apps/$appName/$appName.sh")
         val appScriptSupportTarget = File("$applicationFilesDirPath/${appFilesystem.id}/support/$appName.sh")
-        val appScriptProfileDTarget = File("$applicationFilesDirPath/${appFilesystem.id}/etc/profile.d/$appName.sh")
+        val appScriptProfileDTarget = File("$applicationFilesDirPath/${appFilesystem.id}" +
```
luongthomas: What does D stand for?

luongthomas: In `appScriptProfileDTarget`

corbinlc: The directory where these scripts go is called `profile.d` as in `/etc/profile.d`

luongthomas: Makes sense 👍

PR #324, opened 2018-10-11
Apps polish.

MatthewTighe:
Updates the available filesystem distribution type list dynamically.

Auto-selects correct filesystem in session edit page.

Gets new credentials and starts app session with new filesystem if previous apps filesystem was deleted.

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -20,8 +20,7 @@
     </string-array>
 
     <string-array name="supported_distributions">
-        <item>debian</item>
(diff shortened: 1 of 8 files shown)


app/src/main/res/values/arrays.xml
```diff
@@ -20,8 +20,7 @@
     </string-array>
 
     <string-array name="supported_distributions">
-        <item>debian</item>
-        <item>ubuntu</item>
+        <item>Debian</item>
```
luongthomas: Are we removing ubuntu as a supported distribution for now ?

PR #330, opened 2018-10-12
Production clean up

MatthewTighe:
Cleans up some bugs currently in production.

- Avoids a couple places where `lateinit var`s were uninitialized on reference.
- Start service in foreground earlier to avoid illegal states and get notification running as soon as background work begins.
- Display progress dialog while apps are setting db info up. May fix ANRs.
- Check network reachability before network interactions.

### DIFF
--- a/app/src/main/java/tech/ula/model/repositories/AssetRepository.kt
+++ b/app/src/main/java/tech/ula/model/repositories/AssetRepository.kt
@@ -60,7 +60,7 @@ class AssetRepository(
         val url = "$protocol://github.com/CypherpunkArmory/UserLAnd-Assets-" +
                 "$assetType/raw/master/assets/$architectureType/assets.txt"
 
-        if (!connectionUtility.hostIsReachable(url)) throw object : Exception("Host is unreachable.") {}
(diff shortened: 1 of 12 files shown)


app/src/main/java/tech/ula/model/repositories/AssetRepository.kt
```diff
@@ -59,6 +59,8 @@ class AssetRepository(
 
         val url = "$protocol://github.com/CypherpunkArmory/UserLAnd-Assets-" +
                 "$assetType/raw/master/assets/$architectureType/assets.txt"
+
+        if (!connectionUtility.httpsHostIsReachable("github.com")) throw object : Exception("Host is unreachable.") {}
```
luongthomas: I see "github.com" being used many times, can we put this into strings.xml?

MatthewTighe: strings.xml is meant more for user-facing strings as a list of things that require translation. It would be better to use a constant for sure, but this work will be part of the upcoming refactor anyway. I'm looking to just ship it for today so we can have it stress-tested over the weekend.

PR #331, opened 2018-10-12
Animate apps list

luongthomas:
**Describe the pull request**

Apps list will animate upon refreshing by sliding in from the right.
Also changes AppListFragment from a ListView to RecyclerView and uses DiffUtils to determine list changes.

**Link to relevant issues**

### DIFF
--- a/app/src/main/java/tech/ula/ui/AppListAdapter.kt
+++ b/app/src/main/java/tech/ula/ui/AppListAdapter.kt
@@ -1,11 +1,13 @@
 package tech.ula.ui
 
 import android.app.Activity
+import android.support.constraint.ConstraintLayout
 import android.support.v7.widget.RecyclerView
 import android.view.ContextMenu
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.animation.AnimationUtils
 import android.widget.ImageView
 import android.widget.TextView
 import tech.ula.R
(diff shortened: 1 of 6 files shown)


app/src/main/java/tech/ula/ui/AppListAdapter.kt
```diff
@@ -47,92 +65,177 @@ class AppListAdapter(
...
-        if (convertView == null) {
-            val inflater = activity.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
-            view = when (item) {
-                is AppSeparatorItem -> inflater.inflate(R.layout.list_item_separator, parent, false)
-                is AppItem -> inflater.inflate(R.layout.list_item_app, parent, false)
-            }
+        val diffCallback = AppsItemDiffCallBack(appsAndSeparators, newAppsListItems)
+        val diffResult = DiffUtil.calculateDiff(diffCallback)
+        apps.clear()
+        apps.addAll(newApps)
```
MatthewTighe: Are these two lines still necessary?

app/src/main/java/tech/ula/ui/AppListAdapter.kt
```diff
@@ -47,92 +65,177 @@ class AppListAdapter(
...
-    override fun isEnabled(position: Int): Boolean {
-        return when (appsAndSeparators[position]) {
-            is AppItem -> true
-            is AppSeparatorItem -> false
-        }
+    override fun onViewDetachedFromWindow(holder: ViewHolder) {
+        holder.itemView.clearAnimation()
+    }
+
+    fun getLastSelectedAppsListItem(): AppsListItem {
```
MatthewTighe: Can you rename these functions to indicate that their in relation to context items? Otherwise it seems too much like something that should be handled by the viewmodel.

app/src/main/java/tech/ula/ui/AppListAdapter.kt
MatthewTighe: `apps.clear() and apps.addAll(newApps)` that is

luongthomas: It doesn't look like we're using `apps` at all, it looks safe to remove that list along with these two lines

app/src/main/java/tech/ula/ui/AppListAdapter.kt
luongthomas: How does `lastSelectedAppContextItem` sound?

MatthewTighe: Perfect

PR #345, opened 2018-10-16
Remove paid apps

MatthewTighe:


app/src/main/java/tech/ula/ui/AppListFragment.kt
```diff
@@ -477,34 +447,11 @@ class AppListFragment : Fragment(),
...
-                }
-                .create().show()
-    }
+    private fun userlandIsNewVersion(): Boolean {
+        val info = activityContext.packageManager.getPackageInfo(activityContext.packageName, 0)
+        val version = info.versionName
 
-    override fun onSubscriptionPurchased() {
-        handleAppSelection(lastSelectedApp)
+        val lastUpdatedVersion = activityContext.defaultSharedPreferences.getString("lastAppsUpdate", "")
```
luongthomas: When is `lastAppsUpdate` being set?

PR #391, opened 2018-11-13
Create Unsupported / Broken-Devices List 

EnderNightLord-ChromeBook:
**Describe the pull request**
This will show what devices work with userLAnd and what don't for the time being


corbinlc:
I don't want to commit to creating or maintaining this list.  Maybe we should do the inverse and add a known issues?  Right now I am happy with just adding regular issues if UserLAnd doesn't work for a given device, but I don't think we have found one where it just doesn't work at all.  Are you seeing some devices this does not work for?

corbinlc:
That is the true, the Huawei phones have many issues do to broken POKE_DATA functionality.  I will handle this by creating one summary issue that covers what devices are impacted and what has been done and what the remaining issues are.

EnderNightLord-ChromeBook:
Cool Sounds Good I Will Edit My .txt File

EnderNightLord-ChromeBook:
Does The Huawei 3i Work? Do You Know?

MatthewTighe:
We just don't have the resources to maintain a list like this. Additionally, it can be hard to tell from user reports whether the app is broken because of their device or because of some interaction with the app. This kind of list would also be better placed in the wiki rather than as a somewhat random asset in the repo itself.

Feel free to open an issue requesting this kind of list be added to the wiki if you'd like it to happen eventually, but for now I'm going to close this PR.

PR #393, opened 2018-11-14
Filesystems page will listen to progress bar dialog and display it when navigating to it.

luongthomas:
**Describe the pull request**

When an app or session is starting, the filesystem page elements should not be able to interact with.  This will also eliminate the ability to select on an element when switching to the sessions or apps tab.  (Brief moment where elements are able to be selected before the progress bar appears, e.g. selecting another app)

**Link to relevant issues**


MatthewTighe:
Can the problem of elements being selected be replicated if you navigate from the settings or help page? If so, I don't think this is the right solution. We want to retain as much interactivity for the user as possible.

MatthewTighe:
So I can confirm that this issue is present when navigating from fragments besides the filesystem list. Let's merge this in, as I think it will help and makes sense anyway. We'll need to revisit the root problem soon.

PR #413, opened 2018-11-21
Update help page.

### LINKED ISSUE #412: Help Page partly cut off
Describe the bug

Scrolling down to the bottom of the help page on both available phone rotations the words "behavior" and "linux-like" are still visible, but partly cut off due to the bottom portion of the screen.

To Reproduce

Visit the help page on App version 1.01 beta and use vertical and horizontal landscape/rotation orientation.

Screenshots

[https://imgur.com/a/TGhW5tM](url)

Device Information

Device: [Samsung Galaxy J3 Prime]
Android Version: [7.0]
UserLAnd Version [e.g. 1.0.2]

MatthewTighe:
Closes #412 .

- Add button to navigate to website.
- Move buttons to bottom.
- Remove terms and conditions button.
- Add terms and conditions link to our website to options menu.

image


app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -112,6 +113,10 @@ class MainActivity : AppCompatActivity() {
     }
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        if (item.itemId == R.id.terms_and_conditions) {
+            val intent = Intent("android.intent.action.VIEW", Uri.parse("https://userland.tech/eula"))
```
luongthomas: What do you think of creating an XML file that holds static URLs?  I was thinking this could be abstracted in a way similar to how Strings.xml holds static strings.  Maybe the URLs could even be placed inside strings.xml

MatthewTighe: I think it's adds a layer of unnecessary complexity. Off the top of my head, I can't think of any URLs that are reused by different pieces of code yet. Additionally, it would mean passing a `Resource` object to each method/class that needed to access the URLs.

luongthomas: I was thinking that there's some benefit of being able to look at one location for all strings related to URLs.  I suppose it's not necessary yet since URLs aren't being reused around the codebase.

Issue #414, opened 2018-11-22
How and where can I translate the app?


ahangarha:
Hi

I want to contribute to your project by translating it into Persian. I coudn't find any clue where to start.

corbinlc:
This would be wonderful.  All the text that needs to be translated is found here: https://github.com/CypherpunkArmory/UserLAnd/blob/staging/app/src/main/res/values/strings.xml

There are at least two ways to go about you helping us with the translation:
1) You are familiar with how to use git and build applications with Android Studio:
Then you clone our repo and make the changes yourself and push the changes back to us.  Obviously we will provide any help that is required.  
2) You are not familiar with this things:
You just translate those strings and then we put it in the correct place within our application.

We prefer the first choice, but would take whatever help we can get.  So, do you know how to use git and Android Studio?

ahangarha:
I know git and I can manage to do basic stuff in Android Studio. But I don't know if it is necessary to use Android Studio. I will manage it.

will be back soon :)

corbinlc:
It is not really necessary to use Android Studio, but it would allow you to try out your changes and more easily explore where the strings get used.  I can add the basic directory structure and an empty xml file for you to fill in translations, probably on Monday.  Really appreciate your offer to help.

ahangarha:
I did it :+1:

PR #421, opened 2018-11-23
added Persian translation

ahangarha:
**Describe the pull request**

As discussed in the following issues, I made first translation of the app into Persian. I have tested it. It looks good.

**Link to relevant issues**
#414

### DIFF
--- a/app/src/main/res/values-fa/strings.xml
+++ b/app/src/main/res/values-fa/strings.xml
@@ -19,7 +19,7 @@
     <!-- Menus -->
     <string name="app_info">اطلاعات کاره</string>
     <string name="stop_app">توقف کاره</string>
-    <string name="kill_service">Kill Service</string>
+    <string name="kill_service">توقف خدمت</string>
     <string name="edit">ویرایش</string>
     <string name="delete">پاک‌کردن</string>
     <string name="save">ذخیره‌کردن</string>
@@ -118,7 +118,7 @@
 
     <!-- Permissions necessary alert -->
     <string name="alert_permissions_necessary_title">یوزرلند نیاز به دسترسی دارد!</string>
-    <string name="alert_permissions_necessary_message">To continue, enable storage permissions to allow UserLAnd to download files and move them from external storage to app local storage.</string>
+    <string name="alert_permissions_necessary_message">برای ادامه و برای اینکه یوزرلند بتواند پرونده‌ها را بارگیری و سپس آن‌ها را از حافظه خارجی به حافظه محلی کاره جابجا کند، بایستی اجازه دسترسی به حافظه را فعال کنید.</string>
     <string name="alert_permissions_necessary_cancel_button">لغو</string>
 
     <!-- Network unavailable alert -->


EnderNightLord-ChromeBook:
I Checked... Looks Fine

ahangarha:
Tnx.

Unfortunately I couldn't build the project to check it in real. I don't know if there was some issue with the gradle settings, with internet sanction or with my little knowledge.

But logically there shouldn't be any problem caused by me :)

Hope to see the translation in the app soon

EnderNightLord-ChromeBook:
Yes Good Luck!

app/src/main/res/values-fa/strings.xml
```diff
@@ -0,0 +1,173 @@
...
+    <string name="help">راهنما</string>
+    <string name="toc">شرایط و ضوابط</string>
+
+    <!-- Buttons -->
+    <string name="button_continue">ادامه</string>
+
+    <!-- Menus -->
+    <string name="app_info">اطلاعات کاره</string>
+    <string name="stop_app">توقف کاره</string>
+    <string name="kill_service">Kill Service</string>
```
MatthewTighe: Looks like `Kill Service` still requires translation.

app/src/main/res/values-fa/strings.xml
```diff
@@ -0,0 +1,173 @@
...
+    <!-- Notification strings -->
+    <string name="services_notification_channel_id">خدمات یوزرلند</string>
+    <string name="services_notification_channel_name">خدمات</string>
+    <string name="services_notification_channel_description">خدمات مداوم</string>
+    <string name="service_notification_title">یوزرلند</string>
+    <string name="service_notification_description">یوزرلند در حال اجرای خدمتی در پس‌زمینه است.</string>
+
+    <!-- Progress bar messages -->
+    <string name="progress_bar_start_step">در حال آماده‌سازی برای شما…</string>
+    <string name="progress_basic_app_setup">در حال تنظیم اطلاعات پایگاه داده کاره&#8230;</string>
```
MatthewTighe: `8230#&;` needs to remain together. You have separated it into `8230#&` and `;`. The `;` is an xml special character delimiter. There are several other strings that need this fix as well.

app/src/main/res/values-fa/strings.xml
```diff
@@ -0,0 +1,173 @@
...
+    <string name="alert_play_service_error_message">برای تکمیل این کار، با مشکلی در دسترسی به خدمات گوگل پلی داریم. لطفا دوباره تلاش کنید و در صورت تداوم این مشکل، آن را با توسعه‌دهندگان در میان بگذارید.</string>
+    <string name="alert_play_store_required">کاره‌های پولی برای تایید پرداخت نیازمند دسترسی به خدمات گوگل‌پلی هستند. لطفا برای ادامه، آن را نصب کنید.</string>
+    <string name="alert_subscriptions_unsupported_message">به نظر می‌رسد نسخه گوگل پلی شما برای پشتیبانی اشتراک‌ها، بیش از حد قدیمی است. لطفا برای ادامه، آن را به‌روز کنید.</string>
+    <string name="alert_need_client_app_title">ناتوانی در یافتن کاره کارخواه!</string>
+    <string name="alert_need_client_app_message">برای اجام این کار به یک کارخواه SSH یا VNC نیزا دارد، اما به نظر می‌رسد پلی‌استور نصب نیست. لطفا برای ادامه، یک برنامه کارخواه SSH یا VNC نصب کنید.</string>
+
+    <!-- Permissions necessary alert -->
+    <string name="alert_permissions_necessary_title">یوزرلند نیاز به دسترسی دارد!</string>
+    <string name="alert_permissions_necessary_message">To continue, enable storage permissions to allow UserLAnd to download files and move them from external storage to app local storage.</string>
```
MatthewTighe: Missing translation.

app/src/main/res/values-fa/strings.xml
ahangarha: I left the translation to be done late since it would make no sense if I translate them literally. I need to figure our how else I can translate such terms. I think it can be done later.

app/src/main/res/values-fa/strings.xml
ahangarha: There is no problem. You just see it this way due to mix of RTL and LTR text. If you select the text, you will see there is no `8230#&;` is all together.

app/src/main/res/values-fa/strings.xml
ahangarha: I don't know why I have missed this line. I will make another commit now

app/src/main/res/values-fa/strings.xml
MatthewTighe: Some other phrases in English that communicate the same idea: 
- `Stop Session`
- `Kill Session`
- `End Service`
- `End Session`
- `Terminate Session`
- `Terminate Service`

and other similar phrasing. Let me know if that's helpful or if you'd like more suggestions.

app/src/main/res/values-fa/strings.xml
MatthewTighe: Yep, you're right. My bad!

app/src/main/res/values-fa/strings.xml
```diff
@@ -0,0 +1,173 @@
+<resources>
+    <string name="app_name">یوزرلند</string>
```
luongthomas: Hi @ahangarha,  I'm another developer on the UserLAnd team here.  Thanks for the PR, we appreciate your efforts.  I just wanted to add my comments on the code changes.  What do you think about keeping `"UserLAnd"` untranslated?  (This includes all other lines have `"UserLAnd"` in them, such as line 155)

My reasoning is that regardless of what language their phone is set to, see the app name as `UserLAnd` in the Google Play Store.  Keeping the text in the app to show `"UserLAnd"` will keep the naming consistent in both places.

ahangarha: Hi. I totally agree that name shouldn't be translated. But I haven't translated UseLAnd but I just transliterate it. It is very much common as far as I see. For example Telegram messaging app has Persian title when the phone default language is Persian.

But if you want to keep the text as is, for example because you write it in a very particular way, that is a decision to be made on your side. I will follow what your policy is.

MatthewTighe: Transliteration is probably fine. All we have for checking your work is Google Translate haha, which made us slightly skeptical of the translation as it was coming out as "User Name" or "Usherland".

PR #442, opened 2018-12-03
Apps startup fsm

MatthewTighe:
This PR extricates business logic from the `AppListFragment` into the `AppsStartupFsm` class.

This is the first part of a major refactor that is focused on framing business logic as finite state machines in the model layer, and using a pub-sub paradigm between the view and model layers.


corbinlc:
I want to go over some of this with you.  We can try to do some of it over slack tomorrow, or when you are in the office Wednesday.

PR #479, opened 2018-12-27
Adding built-in terminal

luongthomas:
Using a modified version of Termux, we're adding a built-in terminal that can be used to SSH into the user's UserLAnd filesystem.

There are are a couple more enhancements to be made but the main functionality is there.  If users still want to use other SSH clients such as Connectbot, they are able to.  Upon starting a session, there will be a prompt as to which SSH client they want to use (provided other SSH client apps are installed).

## Files Changed
*File in the termux-app directory are the same as original Termux codebase but with modifications as specified below:

#### Removed Files
Filename | Path
------------ | -------------
README.md | termux-app/
TermuxHelpActivity.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxInstaller.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxOpenReceiver.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxDocumentsProvider.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxFileReceiverActivity.java | termux-app/terminal-term/src/main/java/com/termux/app/

#### Modified Files
Filename | Path
------------ | -------------
settings.gradle | termux-app/
build.gradle | termux-app/terminal-emulator/
TerminalSession.java | termux-app/terminal-emulator/src/main/java/com/termux/terminal/
build.gradle | termux-app/terminal-term/
AndroidManifest.xml | termux-app/terminal-term/src/main/
BackgroundJob.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxActivity.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxService.java | termux-app/terminal-term/src/main/java/com/termux/app/
extra_keys_right.xml | termux-app/terminal-term/src/main/src/main/res/layout/
strings.xml | termux-app/terminal-term/src/main/src/main/res/values/
build.gradle | termux-app/terminal-view

TermuxService.java line 247 is an interesting place.  It shows the intent arguments being passed to dbclient as a startup command.  

Yep.  Created a filesystem in debian, and can use basic bash commands.  Creating a new tab/session in the terminal will also allow access into the same filesystem.  

Then going back to the main UserLAnd page, I'll stop debian and open up ubuntu.  Now the built-in terminal will show a session into the ubuntu filesystem.  

hello_termux


MatthewTighe (changes requested):

app/src/main/java/tech/ula/utils/DownloadUtility.kt
```diff
@@ -52,7 +55,7 @@ class DownloadUtility(
         downloadDirectory.walkBottomUp()
                 .filter { it.name.contains("UserLAnd-") }
                 .forEach {
-                    val delimitedContents = it.name.split("-")
+                    val delimitedContents = it.name.split("-", limit = 3)
```
MatthewTighe: I think this also needs to be changed in `AssetPreferences#getAsset`.

MatthewTighe (changes requested):

termux-app/art/copy-to-other-apps.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
luongthomas: Yeah it's copying launcher icons and we won't use it.  I'll remove this

app/src/main/java/tech/ula/utils/DownloadUtility.kt
luongthomas: The string for `it` is a bit differently formatted than the input for `DownloadUtility`.  For example `libtermux-auth.so-1545942922`

I'm thinking of doing using `split` with a regular expression to separate the string into two strings based on where the last occurrence of the hyphen is.

MatthewTighe: Regardless, I'm pretty sure something will need to be changed in the `AssetPreferences, as `Asset` lists are cached using a string set of the format '<filename>-<remote timestamp>' and they split on that first hyphen.

MatthewTighe (approved):

lucasborer1:
[](url)

Huck3023:
**automaticshare**_application_ `ln`

Huck3023:
#1728

Huck3023:
[huckjonny3923](url)

PR #480, opened 2018-12-27
Architecture refactor v.2.0.0.

MatthewTighe:
This upgrades our architecture to take advantage of state machines in the model layer, which emit `LiveData` events to the `MainActivityViewModel`, which are then consumed and re-emitted as necessary as states to the `MainActivity` if they are pertinent to the user. 

This moves almost all of the code responsible for starting apps and sessions into the model/viewmodel layers, which will allow for a much higher degree of ease during testing. It also separates concerns to adhere more closely to "Clean Architecture".

The way illegal states are displayed needs to be refined, and tests still need to be written for the viewmodel, but that's coming soon.

If you want to start at the top, selection happens in `MainActivity#sessionHasBeenSelected` and `MainActivity#appHasBeenSelected`. If you wanna start at the bottom, `SessionStartupFsm` and `AppsStartupFsm` both have functions that define acceptable state transitions that may be a good place to start. If you wanna start in the middle, `MainActivityViewModel` handles the actual state transition events.

Starting apps and sessions, and all the usual tricks around killing them, trying to start multiple, etc. Hammer on it as hard as you're able

[why_am_i_like_this](https://media.giphy.com/media/l4lR0Q5u91kTNJPsA/giphy.gif)


corbinlc:
Your gif doesn't work.  That is a big issue.

MatthewTighe:
Try again

MatthewTighe:
Alternatively:
image

luongthomas:
When I first start the app it shows the Illegal State notification

app/src/main/java/tech/ula/utils/FilesystemUtility.kt
```diff
@@ -94,14 +97,17 @@ class FilesystemUtility(
...
-        val appScriptProfileDTarget = File("$applicationFilesDirPath/${appFilesystem.id}" +
-                "/etc/profile.d/$fileNameToForceAppScriptToExecuteLast")
-        appScriptSource.copyTo(appScriptSupportTarget, overwrite = true)
+        val appFilesystemProfileDDir = File("$applicationFilesDirPath/${appFilesystem.id}/etc/profile.d")
+        val appScriptProfileDTarget = File("$appFilesystemProfileDDir/$fileNameToForceAppScriptToExecuteLast")
+
+        appFilesystemProfileDDir.mkdirs()
         appScriptSource.copyTo(appScriptProfileDTarget, overwrite = true)
+
+        if (!appScriptProfileDTarget.exists()) throw Exception()
```
luongthomas: Should we use a NoSuchFileException here?

`throw NoSuchFileException(appScriptProfileDTarget)`

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -280,24 +267,16 @@ class DownloadManagerWrapper {
...
-
-class DownloadBroadcastReceiver : BroadcastReceiver() {
-    private var doOnReceived: (Long) -> Unit = {}
-
-    fun setDoOnReceived(action: (Long) -> Unit) {
-        doOnReceived = action
+    fun downloadHasNotFailed(cursor: Cursor): Boolean {
+        if (cursor.moveToFirst()) {
+            val status = cursor.getInt(cursor.getColumnIndex(DownloadManager.COLUMN_STATUS))
+            return status != DownloadManager.STATUS_FAILED
```
luongthomas: I see in `MainActivityViewModel` line 228 that when the state is set to `DownloadsHaveFailed`, that the message the `IllegalState` reason is set to "Failed to download assets."  I think this could be more specific.  When the `DownloadManager`'s status is set to `STATUS_FAILED`, the  `COLUMN_REASON` is updated.  We could use this for the `IllegalState` reason and it may help us better debug why the downloads failed.

[android docs on DownloadManager's COLUMN_REASON property](https://developer.android.com/reference/kotlin/android/app/DownloadManager#COLUMN_REASON%3Akotlin.String)

PR #492, opened 2019-01-02
Hotfix for terminal NPE bug

luongthomas:
An attempt at fixing an issue of Null Pointer Exception when parsing SSH intent to UserLAnd's terminal

There are are a couple more enhancements to be made but the main functionality is there.  If users still want to use other SSH clients such as Connectbot, they are able to.  Upon starting a session, there will be a prompt as to which SSH client they want to use (provided other SSH client apps are installed).

## Files Changed

Filename | Path
------------ | -------------
TermuxActivity.java | termux-app/terminal-term/src/main/java/com/termux/app/
TermuxService.java | termux-app/terminal-term/src/main/java/com/termux/app/

## Where should the reviewer start
TermuxActivity line 300 is where UserLAnd's intent gets received.  It sets an IntentExtra to hold the original intent data as a string.  It gets sent to TermuxService line 150 where it is parsed for hostname, username, port and session name.  

The flow of creating a filesystem and connecting to it via the built in terminal was tested on a Samsung Galaxy S9.

feels_bad_man

### DIFF
--- a/termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
+++ b/termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
@@ -302,7 +302,13 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
             throw new RuntimeException("TermuxActivity dataString from Intent does not have any data");
         }
 
-        serviceIntent.putExtra("intentData", intentData);
+        parseUserlandIntent(intentData);
+
+        serviceIntent.putExtra("username", username);
+        serviceIntent.putExtra("hostname", hostname);
+        serviceIntent.putExtra("port", port);
+        serviceIntent.putExtra("sessionName", sessionName);
+
         serviceIntent.setAction(TermuxService.ACTION_EXECUTE);
 
         // Start the service and make it run regardless of who is bound to it:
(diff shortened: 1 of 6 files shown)


termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -299,8 +298,19 @@ public void onPageSelected(int position) {
         registerForContextMenu(mTerminalView);
 
         Intent serviceIntent = new Intent(this, TermuxService.class);
-        intentData = getIntent().getData();
-        serviceIntent.setData(intentData);
+        String intentData = getIntent().getDataString();
+        if (intentData == null || intentData.isEmpty()) {
+            showErrorAndGoBackToUserland("Intent data is empty.  Only UserLAnd intents are supported");
```
MatthewTighe: This should use a string resource. If you can add it to the resources under UserLAnd and it can be shared with Termux that would be ideal.

termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -313,6 +323,49 @@ public void onPageSelected(int position) {
...
+            }
+
+            if (matcher.find()) {
+                username = matcher.group(1);
+                hostname = matcher.group(2);
+                port = matcher.group(3);
+                sessionName = matcher.group(4);
+            }
+        } catch (Exception e) {
+            showErrorAndGoBackToUserland("Error occurred while parsing intent data with regex: " + e);
```
MatthewTighe: Should also be a resource.

termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -313,6 +323,49 @@ public void onPageSelected(int position) {
...
+        }
+    }
+
+    public void showErrorAndGoBackToUserland(String errorMessage) {
+        AlertDialog.Builder alertDialogBuilder = new AlertDialog.Builder(this);
+
+        alertDialogBuilder.setTitle("An error has occurred")
+                .setMessage(errorMessage)
+                .setCancelable(true)
+                .setPositiveButton("Exit",new DialogInterface.OnClickListener() {
```
MatthewTighe: Same with the "exit".

MatthewTighe: This or something similar should already exist in the UserLAnd resources.

termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
luongthomas: I'll add it as a string resource but keeping it in the module's string.xml resource file may be better.  It'll keep both projects decoupled and allow us to build both terminal and userland projects in parallel.

termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
luongthomas: Good suggestion, I'll add it as a resource

termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
luongthomas: Same as above

PR #506, opened 2019-01-08
Handle terminal notification press

luongthomas:
…

**Describe the pull request**
Fixes the crash related to pressing on the terminal notification.  Rather than start a new terminal activity through an intent, bring the previous terminal activity to the front.  

This will allow the user a one click process to get back to where they previously were in the terminal.

Provide a description of the problem your pull request solves. Screenshots, code examples, etc are welcome in this section.

**Link to relevant issues**

PR #515, opened 2019-01-11
Handle terminal notification press

luongthomas:
**Describe the pull request**

Fixes the crash related to pressing on the terminal notification. Rather than start a new terminal activity through an intent, bring the previous terminal activity to the front.

This will allow the user a one click process to get back to where they previously were in the terminal.

**Link to relevant issues**
#506 
(Current PR has a cleaner git history)

### DIFF
--- a/termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
+++ b/termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
@@ -106,6 +106,10 @@ public final class TermuxActivity extends Activity implements ServiceConnection 
 
     TermuxPreferences mSettings;
 
+    // Don't attempt to unbind from the service unless the client has received some
+    // information about the service's state.
+    private boolean shouldUnbind;
+
     /**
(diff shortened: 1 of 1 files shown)


termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -106,6 +106,10 @@
 
     TermuxPreferences mSettings;
 
+    // Don't attempt to unbind from the service unless the client has received some
+    // information about the service's state.
+    private boolean shouldUnbind;
```
MatthewTighe: Can we explicitly set this to false?

luongthomas: Good catch, we should set it's initial value to false.

MatthewTighe: Primitive booleans in Java are automatically set to false, but I personally think it's better to show that than to expect someone reading it in the future to know that tidbit.

Issue #488, opened 2019-01-01
Cleaning supporting files, without removing filesystems


lapsio:
# Describe the problem your feature request solves.
App is unstable as hell. Downloading assets usually crashes or hangs, sometimes forcing user to Force Stop app in the middle of assets download 

Sometimes it reaults in various app-breaking behavior. Eg I force killed app during downloading assets for new distro (Arch) and now it doesn't re-download them yet installation of new root fails:

# Describe the solution you'd like 
Adding option to clean all files and settings EXCEPT filesystems would be great option saving a lot of trouble.

# Describe alternatives you've considered
I honestly think it's the easiest way to implement this. Alternatively allow to export Filesystems so that they could survive full data clean. Or improve detection of corrupted assets but I believe it could be really hard.

# Additional context

![screenshot_20190101-201531](https://user-images.githubusercontent.com/1054079/50575729-f931ec00-0e03-11e9-8385-71e5af3a6a02.png)

hint: yes I tried. Other filesystems set up correctly. Only Arch is broken but I think it's related to force kill app during assets download.

lapsio:
There's also one more thing to that topic:

![screenshot_20190110-142051](https://user-images.githubusercontent.com/1054079/50995639-f0d26300-151f-11e9-900b-c5a4ff98fa4e.png)

Here's amount of space UserLAnd app takes (3.7G) but when I looked closely at filesystem:

![screenshot_20190110-142051](https://user-images.githubusercontent.com/1054079/50995683-0e073180-1520-11e9-9d4b-dc9eb2c08a60.png)

It really doesn't seem to take above 2.5G. I think it should be possible to get some info about what takes how much space because in current situation there's around 1.5G floating somewhere. And there seem to be no way to clean it without destroying filesystems.

MatthewTighe:
Next release will have this option available in the toolbar menu. See #535

PR #535, opened 2019-01-17
Clear support files option

MatthewTighe:
**Describe the pull request**

This adds a toolbar option to clear all support files. It deletes all directories and assets at the application files directory level. It also deletes all files in filesystem support directories **except** hidden files (like status files) and directories.


app/src/test/java/tech/ula/viewmodel/MainActivityViewModelTest.kt
```diff
@@ -83,7 +86,7 @@ class MainActivityViewModelTest {
         mainActivityViewModel.waitForPermissions(appToContinue = selectedApp)
         mainActivityViewModel.permissionsHaveBeenGranted()
         runBlocking {
-            delay(1)
+            delay(2)
```
luongthomas: We should convert this `2` seconds value to a variable.  

1) It'll add clarity to what it does and 
2) we'll only have to change it in one spot if we need to adjust it in the future.

MatthewTighe:
Closes #488

PR #553, opened 2019-01-24
Add crashlytics logging.

MatthewTighe:
**Describe the pull request**

This adds crashlytics logging. This will allow us to have better snapshots of app state when crashes occur. It is dependent on an api key and build secret which have not been included in the PR.

**Link to relevant issues**
N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,12 +66,12 @@ reference:
   export_fabric_api_key: &export_fabric_api_key
     run:
       name: Export fabric apiKey environment variable
-      command: echo 'apiKey="$FABRIC_API_KEY"' >> app/fabric.properties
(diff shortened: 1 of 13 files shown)


EnderNightLord-ChromeBook:
Thats A Good Idea... How are you going to get the api key?

MatthewTighe:
It will be shared between members of our team.

relan:
FYI, UserLAnd won't get any updates on F-Droid after this change because Crashlytics is proprietary software. https://f-droid.org/en/docs/Inclusion_Policy/

EnderNightLord-ChromeBook:
Why Do You Use F-Droid As Well? Doesn't Most People Use The Playstore?

PR #571, opened 2019-01-29
Fix crash on extracting existing filesystem

MatthewTighe:
**Describe the pull request**
Previously, the app would crash on filesystem extraction if attempting to create a new filesystem of a type that we had already downloaded assets for. Normally, these (already downloaded) assets should just be copied to the new filesystem, but there was a bug in the flow such that assets were only copied to filesystems immediately after download.

This PR moves the "verification" step, which is now responsible for copying assets to filesystems, from after extraction to before it, to ensure that all assets are present before extraction step.

**Link to relevant issues**


app/src/main/java/tech/ula/model/state/SessionStartupFsm.kt
```diff
@@ -198,79 +198,67 @@ class SessionStartupFsm(
...
+
+            try {
+                filesystemUtility.copyAssetsToFilesystem("${filesystem.id}", filesystem.distributionType)
+                filesystem.lastUpdated = timeUtility.getCurrentTimeMillis()
+                filesystemDao.updateFilesystem(filesystem)
+            } catch (err: Exception) {
+                state.postValue(FilesystemAssetCopyFailed)
+            }
+
+            val filesystemDirectoryName = "${filesystem.id}"
```
luongthomas: Does this variable need to be redeclared?  There is another variable with the same name in line 216

MatthewTighe: It does not need to be redeclared, no.

Issue #540, opened 2019-01-18
Feature Request - Add native XSDL support


ankushlohiya001:
XSDL sever is working all fine, but can you please add xsdl inside the userland, as you already added a terminal,,
ie.
Same that we get in debian noroot,,

Marcus0086:
Source code is available. So if u r a SE then add it urselves.

corbinlc:
@ankushlohiya001, that is on our roadmap.  Can you please update your feature request to follow the provided template. 
@Marcus0086, it is ok for people to make feature requests.  I think people know they can contribute.

Marcus0086:
Hey hey @corbinlc so when Fedora 28 or 29 is coming

MatthewTighe:
@Marcus0086 Please keep conversation relevant to posted issues.

EnderNightLord-ChromeBook:
this would be posslible... I have seen people add different apps and smashed them into 1... so the other app is an asset.

EnderNightLord-ChromeBook:
I will do some research on this... and check back in with you guys.

lewisdonofrio:
No need to add it to this project, the two projects are both actively updating on my note8, just my 2cents.

PR #578, opened 2019-01-31
Add initial XSDL support

luongthomas:
**Describe the pull request**

Adds an option to connect via XSDL when creating a filesystem or starting an app for the first time.

There is still additional work to be done including checking when the XSDL server is running via the normal flow (ServerUtility.kt line115) but will be added at a later date.

screenshot_20190129-161206_xserver xsdl

**Link to relevant issues**
 #540

### DIFF
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -37,7 +37,6 @@ import kotlinx.android.synthetic.main.activity_main.* // ktlint-disable no-wildc
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.selects.select
 import org.jetbrains.anko.defaultSharedPreferences
 import org.jetbrains.anko.find
 import tech.ula.model.entities.App
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -37,6 +37,7 @@ import kotlinx.android.synthetic.main.activity_main.* // ktlint-disable no-wildc
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.selects.select
 import org.jetbrains.anko.defaultSharedPreferences
 import org.jetbrains.anko.find
 import tech.ula.model.entities.App
@@ -565,8 +566,12 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
             customDialog.getButton(AlertDialog.BUTTON_POSITIVE).setOnClickListener {
                 customDialog.dismiss()
                 val sshTypePreference = customDialog.find<RadioButton>(R.id.ssh_radio_button)
-                val selectedPreference =
-                        if (sshTypePreference.isChecked) SshTypePreference else VncTypePreference
+                val vncTypePreference = customDialog.find<RadioButton>(R.id.vnc_radio_button)
+                val selectedPreference = when {
+                    sshTypePreference.isChecked -> SshTypePreference
+                    vncTypePreference.isChecked -> VncTypePreference
+                    else -> XsdlTypePreference
+                }
                 viewModel.submitAppServicePreference(selectedPreference)
             }
         }
--- a/app/src/main/java/tech/ula/ServerService.kt
+++ b/app/src/main/java/tech/ula/ServerService.kt
@@ -139,6 +139,7 @@ class ServerService : Service() {
         when (session.serviceType) {
             "ssh" -> startSshClient(session, "org.connectbot")
             "vnc" -> startVncClient(session, "com.iiordanov.freebVNC")
+            "xsdl" -> startXsdlClient(session, "x.org.server")
             else -> sendDialogBroadcast("unhandledSessionServiceType")
         }
         sendSessionActivatedBroadcast()
@@ -162,6 +163,17 @@ class ServerService : Service() {
 
         if (clientIsPresent(bVncIntent)) {
             this.startActivity(bVncIntent)
+        } else {
+            getClient(packageName)
+        }
+    }
+
+    private fun startXsdlClient(session: Session, packageName: String) {
+        val xsdlIntent = Intent(Intent.ACTION_VIEW)
+        xsdlIntent.data = Uri.parse("x11://give.me.display:4721")
+
+        if (clientIsPresent(xsdlIntent)) {
+            startActivity(xsdlIntent)
         } else {
             getClient(packageName)
         }
--- a/app/src/main/java/tech/ula/utils/AndroidUtility.kt
+++ b/app/src/main/java/tech/ula/utils/AndroidUtility.kt
@@ -140,12 +140,19 @@ object VncTypePreference : AppServiceTypePreference() {
     }
 }
 
+object XsdlTypePreference : AppServiceTypePreference() {
+    override fun toString(): String {
+        return "xsdl"
+    }
+}
+
 class AppsPreferences(private val prefs: SharedPreferences) {
 
     fun setAppServiceTypePreference(appName: String, serviceType: AppServiceTypePreference) {
         val prefAsString = when (serviceType) {
             is SshTypePreference -> "ssh"
             is VncTypePreference -> "vnc"
+            is XsdlTypePreference -> "xsdl"
             else -> "unselected"
         }
         with(prefs.edit()) {
@@ -160,6 +167,7 @@ class AppsPreferences(private val prefs: SharedPreferences) {
         return when {
             pref.toLowerCase() == "ssh" || (app.supportsCli && !app.supportsGui) -> SshTypePreference
             pref.toLowerCase() == "vnc" || (!app.supportsCli && app.supportsGui) -> VncTypePreference
+            pref.toLowerCase() == "xsdl" || (!app.supportsCli && app.supportsGui) -> XsdlTypePreference
             else -> PreferenceHasNotBeenSelected
         }
     }
--- a/app/src/main/java/tech/ula/utils/ServerUtility.kt
+++ b/app/src/main/java/tech/ula/utils/ServerUtility.kt
@@ -85,6 +85,7 @@ class ServerUtility(
         val command = "../support/execInProot.sh /bin/bash -c '/support/startXSDLServer.sh'"
         return try {
             val env = HashMap<String, String>()
+            env["INITIAL_USERNAME"] = session.username
             env["DISPLAY"] = ":4721"
             env["PULSE_SERVER"] = "127.0.0.1:4721"
             val process = execUtility.wrapWithBusyboxAndExecute(targetDirectoryName, command, doWait = false, environmentVars = env)
--- a/app/src/main/java/tech/ula/utils/ServerUtility.kt
+++ b/app/src/main/java/tech/ula/utils/ServerUtility.kt
@@ -1,8 +1,5 @@
 package tech.ula.utils
 
-import android.content.Intent
-import android.support.v4.content.LocalBroadcastManager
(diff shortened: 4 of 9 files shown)


app/src/main/java/tech/ula/utils/ServerUtility.kt
```diff
@@ -79,6 +79,23 @@ class ServerUtility(
         }
     }
 
+    private fun startXSDLServer(session: Session): Long {
```
MatthewTighe: I think this function, and maybe the asset names could be changed to better reflect what they're doing. It doesn't seem like we're really starting a 'server' per se.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -160,6 +167,7 @@ class AppsPreferences(private val prefs: SharedPreferences) {
         return when {
             pref.toLowerCase() == "ssh" || (app.supportsCli && !app.supportsGui) -> SshTypePreference
             pref.toLowerCase() == "vnc" || (!app.supportsCli && app.supportsGui) -> VncTypePreference
+            pref.toLowerCase() == "xsdl" || (!app.supportsCli && app.supportsGui) -> XsdlTypePreference
```
MatthewTighe: The second half of the conditional here will never be hit, as it would already be true for the expression in the line above.

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -565,8 +565,12 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
             customDialog.getButton(AlertDialog.BUTTON_POSITIVE).setOnClickListener {
                 customDialog.dismiss()
                 val sshTypePreference = customDialog.find<RadioButton>(R.id.ssh_radio_button)
-                val selectedPreference =
-                        if (sshTypePreference.isChecked) SshTypePreference else VncTypePreference
+                val vncTypePreference = customDialog.find<RadioButton>(R.id.vnc_radio_button)
```
MatthewTighe: I think it's worth being explicit here, instead of just using an `else` branch to default to Xsdl. I don't know if the `else` branch could ever be hit there, but just in case, error checking would then need to be added to `AppsStartupFsm#setAppServicePreference`.

luongthomas:
Once https://github.com/CypherpunkArmory/UserLAnd-Assets-Support/pull/91 is pulled in, testing can be readily done on your device.  

Until then, two changes can be made to make it work:
1. AssetRepository.kt 89 - `branch = "add-xsdl"`
2. DownloadUtility.kt 30 - `branch = "add-xsdl"`

luongthomas:
Disabled option for XSDL if running Android 9.0

screenshot_20190201-105809_userland

screenshot_20190201-110854_userland

app/src/main/java/tech/ula/MainActivity.kt
luongthomas: I agree, I've added another variable for xsdlTypePreference here to be more explicit

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I've removed the second half of the conditional, I can understand why it would not be hit there

app/src/main/java/tech/ula/utils/ServerUtility.kt
luongthomas: I've changed it to `setDisplayNumberAndStartTwm` to more accurately describe what the startup script is doing.

app/src/main/java/tech/ula/ServerService.kt
```diff
@@ -167,6 +168,17 @@ class ServerService : Service() {
         }
     }
 
+    private fun startXsdlClient(session: Session, packageName: String) {
+        val xsdlIntent = Intent(Intent.ACTION_VIEW)
+        xsdlIntent.data = Uri.parse("x11://give.me.display:4721")
```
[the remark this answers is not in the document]
luongthomas: You're right, I've changed it to Intent.FLAG_ACTIVITY_NEW_TASK

app/src/main/java/tech/ula/utils/AndroidUtility.kt
AndrewScibek: Wouldn't this never hit the third condition?

luongthomas: If you're referring to `app.supportsGui` as the third condition then yes it would never hit that part.  So I removed that part too

[The line is now reduced to](https://github.com/CypherpunkArmory/UserLAnd/pull/578/files#diff-97f43415e5835dd28a9d0d0c0387503eR170):
`pref.toLowerCase() == "xsdl" -> XsdlTypePreference`

EnderNightLord-ChromeBook:
hey @luongthomas? Why does android 9.0 not support XSDL. I'm Just Curious.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
MatthewTighe: Oh, @AndrewScibek  is actually mostly right. The third condition is only ever hit if `prefs.toLowerCase() == "vnc"` is false *and* (!app.supportsCli && app.supportsGui)` is false. 

I think you could probably leave out your conditional completely. The expression is meant to allow us to default to either SSH if we don't support a graphical environment for an app, or VNC if we don't support a terminal for it. We'd need some kind of top-level user setting to determine if they wanted stuff like this to default to VNC or X.

luongthomas:
> hey @luongthomas? Why does android 9.0 not support XSDL. I'm Just Curious.

There's a known issue with XSDL on devices for version 9.0 where the app just restarts endlessly after 5 seconds.  
https://github.com/pelya/xserver-xsdl/issues/62

app/src/main/java/tech/ula/utils/AndroidUtility.kt
MatthewTighe: Additionally, given that XSDL doesn't have support for Android 9 right now, I think the safe bet is to just default to VNC. Saves the overhead of doing extra version guarding there as well.

EnderNightLord-ChromeBook:
> There's a known issue with XSDL on devices for version 9.0 where the app just restarts endlessly after 5 seconds.

Thanks @luongthomas

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -220,25 +222,62 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
...
+    private fun setXsdlDisplay() {
+        try {
+            val i = Intent(Intent.ACTION_MAIN, Uri.parse("x11://give.me.display:4721"))
+            startActivityForResult(i, 1)
+        } catch (e: Exception) {
+            val appPackageName = "x.org.server"
+            try {
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=$appPackageName")))
+            } catch (error: android.content.ActivityNotFoundException) {
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse("https://play.google.com/store/apps/details?id=$appPackageName")))
```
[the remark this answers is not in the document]
luongthomas: On our Nexus test device (without the Play store installed) it sends it to the webpage for XSDL rather than throwing an exception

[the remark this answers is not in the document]
MatthewTighe: Oh, duh, it's a url intent 😅

app/src/main/res/layout/frag_app_details.xml
```diff
@@ -83,5 +83,18 @@
 
         </RadioGroup>
 
+        <TextView
```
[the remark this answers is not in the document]
luongthomas: Since it's a scroll view everything is visible, but I've added some padding to the bottom so that this text isn't clipped halfway anymore.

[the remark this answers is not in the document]
luongthomas: On the Nexus S (4 inches for the display size, measuring diagonally on the screen) The dia_app_select_client.xml was chopping off the xsdl version support text.  Which I think turning into a scrollview would be best

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -220,25 +222,62 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
...
 
-    private fun startSession(session: Session) {
+    private fun prepareSessionForStart(session: Session) {
         val step = getString(R.string.progress_starting)
         val details = ""
         updateProgressBar(step, details)
 
+        if (session.serviceType == "xsdl" && Build.VERSION.SDK_INT >= 27) {
+            session.serviceType = "vnc"
+            startSession(session)
```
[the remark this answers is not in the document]
luongthomas: I've added a trello card and a TODO comment

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -220,25 +222,62 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
...
+    private fun startSession(session: Session) {
         val serviceIntent = Intent(this, ServerService::class.java)
                 .putExtra("type", "start")
                 .putExtra("session", session)
         startService(serviceIntent)
     }
 
+    private fun setXsdlDisplay() {
+        try {
+            val i = Intent(Intent.ACTION_MAIN, Uri.parse("x11://give.me.display:4721"))
```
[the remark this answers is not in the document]
luongthomas: Yep sure

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -220,25 +222,62 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
...
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=$appPackageName")))
+            } catch (error: android.content.ActivityNotFoundException) {
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse("https://play.google.com/store/apps/details?id=$appPackageName")))
+            }
+        }
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(requestCode, resultCode, data)
+        if (data != null) {
```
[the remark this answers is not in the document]
luongthomas: I've changed it to data?.let for consistency

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -166,7 +166,7 @@ class AppsPreferences(private val prefs: SharedPreferences) {
 
         return when {
             pref.toLowerCase() == "ssh" || (app.supportsCli && !app.supportsGui) -> SshTypePreference
-            pref.toLowerCase() == "vnc" || (!app.supportsCli && app.supportsGui) -> VncTypePreference
+            pref.toLowerCase() == "vnc" -> VncTypePreference
```
MatthewTighe: I believe the `||` branch in the `vnc` conditional expression is still necessary to get the defaulting behavior we want.

EnderNightLord-ChromeBook: Yeah!!! This Update Will Be Awesome When It's Done. I Can't Wait

luongthomas: What do you think about this ordering of conditionals?

```
return when {
    pref.toLowerCase() == "ssh" || (app.supportsCli && !app.supportsGui) -> SshTypePreference
    pref.toLowerCase() == "xsdl" -> XsdlTypePreference
    pref.toLowerCase() == "vnc" || (!app.supportsCli && app.supportsGui) -> VncTypePreference
    else -> PreferenceHasNotBeenSelected
}
```

If the app already has XSDL set as the client type, the [appDetail] page will reflect that.  Otherwise if it's not set to SSH or VNC, then it will default to VNC if the app doesn't support CLI but supports GUI.

Issue #500, opened 2019-01-07
I want to improve Spanish translation. Where to start?


MatiasMFM2001:
This app is simply great. I'd like to help programming it, but I don't have enough experience. So, I can help you at least improving Spanish translation. I've searched "Spanish" and "Language" in your Github repo (code section) with no results.

What exactly files do I need to modify, in order to create a Pull Request and update the translation?

corbinlc:
Please see #430 with regards to how to get started and what files are of interest.  If you are using the beta version of the app, you will see that there is a Spanish translation in place (the file related to that is in the values_es directory).  That translation was added last week by my 8yr old daughter who is bilingual, so definitely curious how good it is.  Some of the those translations are not exact translations as she also simplified a couple of phrases.

MatiasMFM2001:
Thank you!

Yes, her translation is good! I think her Spanish is better than my English...

I'll work in the translation!

MatiasMFM2001:
I've finished! BUT I'm noob in GitHub, so I'll just copy translated strings.txt here. Thank you for your great work!!

``` 
﻿<?xml version="1.0" encoding="utf-8"?>
<resources>
    <string name="app_name">UserLAnd</string>
    <string name="welcome">¡Bienvenido a UserLAnd!</string>
    <string name="permission_request_code">1000</string>
    <string name="apps">Aplicaciones</string>
    <string name="error_session_name">Tienes que especificar un nombre de sesión.</string>
    <string name="error_empty_field">¡Debes completar todos los campos!</string>
    <string name="error_filesystem_name">Tienes que especificar un nombre para el sistema de archivos.</string>
    <string name="error_username">Tienes que especificar un nombre de usuario.</string>
    <string name="sessions">Sesiones</string>
    <string name="filesystems">Sistemas de Archivos</string>
    <string name="edit_session">Editar Sesión</string>
    <string name="edit_filesystem">Editar Sistema de Archivos</string>
    <string name="settings">Configuración</string>
    <string name="help">Ayuda</string>
    <string name="button_continue">Continuar</string>
    <string name="hint_username">Nombre de Usuario</string>
    <string name="hint_session_name">Nombre de Sesión</string>
    <string name="hint_filesystem_name">Nombre de Sistema de Archivos</string>
    <string name="refresh">Actualizar</string>
    <string name="free_annotation">(Gratis)</string>
    <string name="paid_annotation">(Pago)</string>
[...]

PR #494, opened 2019-01-03
I made a Spanish translation for your app

dragonSilverString:
**Describe the pull request**

I made the translation so that people who only speak Spanish or prefer Spanish could use my Daddy’s app.

**Link to relevant issues**

#430

Issue #430, opened 2018-11-27
Translations Needed


corbinlc:
**Is your feature request related to a problem? Please describe.**
Many of our users are non English speakers, but like many apps we have only supported English.  As of today we now have translations for Chinese and Persian added to our repo, but there are so many more languages that would help.  For example, we have more installs in Brazil than any other country, but we don't have a Portuguese translation.  

**Describe the solution you'd like**
I would like anyone to get involved and start contributing translations for any language that they are fluent in.  Doing so is fairly easy as most of the text is found in [one file](https://github.com/CypherpunkArmory/UserLAnd/blob/staging/app/src/main/res/values/strings.xml) but we will help you as much as needed to make this happen.

**Describe alternatives you've considered**
We don't want to pay for this service because we want someone who is technical and part of our community to partner with us in bringing UserLAnd to your native tongue.

AuntieE:
Hi,
   I am a native English speaker, but work as a member/admin of the French translation team of Full Circle Magazine (https://fullcirclemagazine.org ; www.fullcirclemag.fr) devoted to Linux Ubuntu in all its forms and, more particularly to Ubuntu.
  Have I understood correctly that what you need is a translation of the commands in your file? That would be possible I think. Just let me know if you think I can help. In a former life, I taught translation from French to English to graduate students at the Sorbonne. Translation theory is the same in the other direction...

dragonSilverString:
I think you are perfectly qualified for translating the string.xml to french given your job, your skills and your experience using our app.  We would love to have you take a crack at it.

AuntieE:
OK! Will do my best.

Provenance : Courrier<https://go.microsoft.com/fwlink/?LinkId=550986> pour Windows 10

________________________________
De : dragonSilverString <notifications@github.com>
Envoyé : Wednesday, November 28, 2018 6:20:16 PM
À : CypherpunkArmory/UserLAnd
Cc : AuntieE; Manual
Objet : Re: [CypherpunkArmory/UserLAnd] Translations Needed (#430)


I think you are perfectly qualified for translating the string.xml to french given your job, your skills and your experience using our app. We would love to have you take a crack at it.

corbinlc:
I think you are perfectly qualified for translating the string.xml to french given your job, your skills and your experience using our app. We would love to have you take a crack at it.

corbinlc:
Sorry, my daughters github account was logged in.  Made the same comment now using my name.

[...]

PR #582, opened 2019-02-02
Some comments added to English strings.xml

MatiasMFM2001:
**Describe the pull request**

I've added some comments so @corbinlc won't need to explain other translators what he has explained to me. Maybe you should add other comments!

**Link to relevant issues**
#500 #430 #494

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,10 +1,10 @@
-<resources>
+<resources>                                                         <!-- Translators, pay attention to this notes! --> 
     <string name="app_name">UserLAnd</string>
     <string name="welcome">Welcome to UserLAnd!</string>
     <string name="permission_request_code">1000</string>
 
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1,10 +1,10 @@
-<resources>
+<resources>                                                         <!-- Translators, pay attention to this notes! -->
```
EnderNightLord-ChromeBook: pay attention to this notes

hows about 

pay attention to these notes

MatiasMFM2001: You're right! Thanks for marking that!

EnderNightLord-ChromeBook: your welcome

EnderNightLord-ChromeBook: And Everything Else Looks Good.

Issue #561, opened 2019-01-26
Download problems


18227077056:
1.Sometimes it will download 45/27 files or more
2.Can not download
3.It works slowly
4..........

EnderNightLord-ChromeBook:
Please stick to the issue template or we will have to close this.

Marcus0086:
"https://github.com/CypherpunkArmory/UserLAnd/issues/new?template=bug_report.md&title=Bug+report+-+%5BYOUR+BUG+TITLE%5D" use it from here

corbinlc:
@18227077056 I am one of the developers and it is true that we want people to provide info as shown in the template.  After you provide the needed info, we will respond with more questions/comments that should help work towards a resolution of your problem.  Thanks in advance!

Issue #559, opened 2019-01-25
Nothing will download


T4O2M0:
Describe the bug
I open the app, tap any linux distro or app to download and its stuck at downloading assets forever

Go to 'apps'

Click on 'any app or distro'

See error

Device Information
Device: [LG G7]
Android Version: [8.0]
UserLAnd Version [2.2.1]
Not rooted
Not running lineageOS

Additional context
Download manager shows the downloads "queued"

T4O2M0:
So you just gonna tell me to update it and then not help me?

EnderNightLord-ChromeBook:
This is to get the best information. If you want help just stick to the template and we can help you.

Marcus0086:
Hey hey @T4O2M0 just unistall userland and then reinstall it. Now clickbon any distro and let it download all the assests and also let it to unzip them. Remember during the downloading and unzipping don't click on screen just let it as it is. Don't do anything on screen or u just increase the screen off time to 30 min. If u got a descent internet of 300-500kbps then it will download in no time.

Marcus0086:
Now just close the issue.

T4O2M0:
@Marcus0086 still having the same results :( could it be my phone?

Marcus0086:
Hmm @T4O2M0 maybe. But wait for tomorrow bcz my friend got the same phone as u mentioned in that comment. So let me first check it. Then i will be able to help you. Thank you

Marcus0086:
Hey @T4O2M0 do you have any other descent smartphone in your house. If yes, so then install userland on it and observe if anything happens.

T4O2M0:
@Marcus0086 The only other phone I have is an iPhobene 5c :/ although I do have a android tablet thats semi old, should I try that?

Marcus0086:
Hmm are specs not descent enough

T4O2M0:
My lg g7 should be more than decent enough but im not sure about the tablet.

Marcus0086:
Then give it a try.

corbinlc:
I have not heard of issues with that device.  @MatthewTighe is the best one to help debug downloading problems.  Now that you have updated your issue, he will probably respond in the next couple of days with some questions to help resolve this.

MatthewTighe:
@T4O2M0 Does the download manager work normally? The download manager is an Android framework component. Once we queue downloads with it, we have no control over those downloads until they are completed. If they are being queued and not making progress, there's nothing we can do about that.

[...]

Issue #538, opened 2019-01-18
Illegal State - Bad transition at the end of assets download


lapsio:
# Describe the bug

Error after assets download

# Steps to reproduce the behavior:

Clear Support Files
start session

# Screenshots

![screenshot_20190118-023546](https://user-images.githubusercontent.com/1054079/51360133-b42bdc00-1aca-11e9-9d4e-a20ea40abc2d.png)

# Device Information

Device: ASUS Zenphone 3 ()
Android Version: Oreo, 8.0
UserLAnd Version 2.2.0
not rooted
stock ROM

# Additional context

Filesystem seems to be functional but I'm not sure if it's serious.

MatthewTighe:
Please use the bug report template to update your original post. Reproducible steps, device information, etc is still useful in resolving this issue.

The download process does use Android standard download manager. If you aren't seeing the downloads being down through it, I can only assume you're using a non-standard system.

lapsio:
> Please use the bug report template to update your original post. Reproducible steps, device information, etc is still useful in resolving this issue.
> 
> The download process does use Android standard download manager. If you aren't seeing the downloads being down through it, I can only assume you're using a non-standard system.

Updated. I'll post info about download process to thread I previously created on this topi

MatthewTighe:
I'm having trouble reproducing this issue. You listed every step necessary to reproduce it on your device?

lapsio:
I think It's related to files download problem. There's high chance not all files have been downloaded and launcher proceeded despite not having all necessary assets.

MatthewTighe:
Are you unable to reproduce this now? I'd rather leave it open if it's still an issue.

lapsio:
Yes, If files are improperly downloaded then there's this error. However I guess as filesystem is old (and fully created already) UserLAnd still manages to launch ssh to it.

lapsio:
Also those errors vary. I guess it depends on how many assets have actually been downloaded and how many are missing.

MatthewTighe:
@lapsio Are you on the beta channel? I'm thinking of switching to tracking downloads using their ID which should be more robust.

MatthewTighe:
Okay, I'm pushing a beta release which might help with downloads a bit. Note that this release will cause crashes after it displays the Illegal State alert. This is so we can gather stack traces from illegal states, and the crash will be removed next week when we move the release to production.

[...]

PR #590, opened 2019-02-05
Add download state cache.

MatthewTighe:
**Describe the pull request**
We were experiencing crashes for illegal state when asset downloads were completed, due to the state of the session FSM being incorrect, most likely due to process death and re-initialization. This PR adds logic to cache download state as it's occuring, and then the session fsm syncs with the cache when it is created. Most of the logic surrounding downloads has also been moved into the download utility itself.

A good place to start would be in `SessionStartupFSM#handleAssetsDownloadComplete`.

**Link to relevant issues**
#561 #559 #538

### DIFF
--- a/app/src/main/java/tech/ula/model/state/SessionStartupFsm.kt
+++ b/app/src/main/java/tech/ula/model/state/SessionStartupFsm.kt
@@ -175,6 +175,8 @@ class SessionStartupFsm(
     }
 
     private fun handleAssetsDownloadComplete(downloadId: Long) {
+        if(!downloadUtility.downloadIsForUserland(downloadId)) return
+
         if (!downloadUtility.downloadedSuccessfully(downloadId)) {
             val reason = downloadUtility.getReasonForDownloadFailure(downloadId)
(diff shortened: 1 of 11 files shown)


app/src/main/java/tech/ula/model/state/SessionStartupFsm.kt
```diff
@@ -164,38 +166,28 @@ class SessionStartupFsm(
...
+    }
 
-        downloadedIds.sort()
-        downloadingIds.sort()
-        if (downloadedIds != downloadingIds) {
-            state.postValue(DownloadsHaveFailed("Downloads completed with non-enqueued downloads"))
-            return
+    private fun handleAssetDownloadState(assetDownloadState: AssetDownloadState) {
+        return when (assetDownloadState) {
+            // We don't care if we've something else has downloaded something.
```
luongthomas: Looks like there's an extra word somewhere in this line

MatthewTighe: I grammar goodly

Issue #577, opened 2019-01-31
Wrong password : Arch


XeGrox:
When I enter the password it is always wrong, although I am very sure I entered the correct password and even tried recreating the file system with a different password 3 times. I have also tried 'userland'...

This seems to only affect arch

corbinlc:
In general Arch works for people, so we are going to need some more info to understand what is going wrong.  To start, can you update your issue to follow this template...

https://github.com/CypherpunkArmory/UserLAnd/issues/new?template=bug_report.md&title=Bug+report+-+%5BYOUR+BUG+TITLE%5D

That template includes some info that we often find helpful when debugging.

XeGrox:
Create a new app and fill in the password blank.
Attempt to login with the same password in ssh login
SSH prompts your password again and again no matter what you type

Emui 9.0 (android 9.0)
Userland v2.2.3

I have tried reinstalling, clearing support files and clearing data but to no avail

Seems to only affect the arch distrubution

MatthewTighe:
I can't reproduce this. If you aren't entering your password incorrectly (which doesn't seem to be the case), my only guess is that the password you're trying to use is somehow not formed correctly. This could potentially be something like weird characters, or a length that isn't suitable. If that were the case, we can update our validation of ssh passwords to guard against it.

@corbinlc Any insight into this?

EnderNightLord-ChromeBook:
did you try uninstalling? Did you try older versions?

XeGrox:
@MatthewTighe Yep I have tried using numbers/letters/longer/shorter but it still fails

@EnderNightLord-ChromeBook I don't think older versions would help since the file system is downloaded online but I have tried reinstalling.

corbinlc:
@XeGrox the only interesting thing I can think of is that you are one of the first users of EMUI 9.  EMUI 9 solves a lot of issues we had with Huawei devices previously, but maybe there are still some oddities that are Huawei specific.  Can you try to capture a proot log of you trying to log in?  Here is how to do that...
1) stop any running app/distro
2) in the app settings, enable proot debugging and turn the proot debug verbosity to 9
3) start arch distro again and try to log in a couple of times and seeing it fail
4) send the file, normally found at /sdcard/PRoot_Debug_Log to us
There should be some interesting breadcrumb in that log.
Thanks,
Corbin

MatthewTighe:
[...]

Issue #460, opened 2018-12-07
Password set in new app password dialog not working


lmorda:
When creating a new app, the Password that I entered into the please enter the relevent required information dialog during the setup process does not appear to be working when I try to SSH into the Linux image.

I created a Debian app on my Google Pixel XL running Android 8.0.  Everything worked correctly.  As soon as the app launched, JuiceSSH was brought up to the foreground as expected.  JuiceSSH then asked for the password for the username I created.  I entered the password and the login failed.  Tried this whole process multiple times and still experiencing the same issue.  Also tried other SSH clients and still experiencing the failed password problem.

I also tried entering the default password "userland" with no success.

corbinlc:
We have not heard of this issue.  Is it possible that you entered your password incorrectly when originally creating the filesystem (we should probably make the form ask you to confirm what you entered)?  Can you try again from scratch and tell us how it goes?  If that does not work...is there something special about your password, like special characters.

lmorda:
Yeah no problem tried again three more times and same problem.  Nothing special about password and not sensitive info:

Username
operator
Password
operator
VNC Password
operator

I'm on a Google Pixel XL running Android 8.0 but not sure that makes a difference.  I could try another device too when I get into work on Monday.

corbinlc:
that particular username and password do not work for me either (all others i have ever used have worked).  maybe the username and password cannot be the same.  let me see what the rules are

corbinlc:
`useradd: group operator exists - if you want to add this user to that group, use -g.`

So it looks like that is already an existing group and by default a new user is created with a group name that matches their username.  

So, you cannot user `operator` as your username.  I will add an internal card for improving our form validation.

corbinlc:
I assume that will solve your issue.  I am going to close this, but please reopen it if this specific issue is not resolved.

lmorda:
I also saw this error:

Extracting: chsh: user 'operator' does not exist

lmorda:
Ok perfect I should have tried that for you !  Thanks

PR #597, opened 2019-02-07
Refactor filesystem credential validation

luongthomas:
**Describe the pull request**

Moved username and password validation logic from `MainActivity` into the `ValidationUtility` class.  Now it will return a `Credential` object that has two properties: a boolean of whether the string is valid and an error message to display in a Toast dialog about why the string is invalid.

Also added a blacklist of usernames that the user cannot choose from because they are used in Linux distros as group names.

There are now some tests too 👍 

**Link to relevant issues**
#460
#577

### DIFF
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -659,21 +659,27 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
     }
 
     private fun validateCredentials(username: String, password: String, vncPassword: String): Boolean {
-        val resources = this.resources
-        val validator = ValidationUtility(resources)
-        var allCredentialsAreValid = false
+        val blacklistedUsernames = this.resources.getStringArray(R.array.blacklisted_usernames)
+        val validator = ValidationUtility(blacklistedUsernames)
 
         val usernameCredentials = validator.validateUsername(username)
-        val passwordCredentials = validator.validatePasswords(password, vncPassword)
+        val passwordCredentials = validator.validatePassword(password)
+        val vncPasswordCredentials = validator.validateVncPassword(vncPassword)
 
-        when {
-            !usernameCredentials.credentialIsValid ->
-                Toast.makeText(this, usernameCredentials.errorMessage, Toast.LENGTH_LONG).show()
-            !passwordCredentials.credentialIsValid ->
-                Toast.makeText(this, passwordCredentials.errorMessage, Toast.LENGTH_LONG).show()
-            else ->
-                allCredentialsAreValid = true
+        return when {
+            !usernameCredentials.credentialIsValid -> {
+                Toast.makeText(this, usernameCredentials.errorMessageId, Toast.LENGTH_LONG).show()
+                false
+            }
+            !passwordCredentials.credentialIsValid -> {
+                Toast.makeText(this, passwordCredentials.errorMessageId, Toast.LENGTH_LONG).show()
+                false
+            }
+            !vncPasswordCredentials.credentialIsValid -> {
+                Toast.makeText(this, passwordCredentials.errorMessageId, Toast.LENGTH_LONG).show()
+                false
+            }
+            else -> true
         }
-        return allCredentialsAreValid
     }
 }
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -658,31 +658,21 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
         customDialog.show()
     }
 
-    // TODO the view shouldn't be responsible for validation
     private fun validateCredentials(username: String, password: String, vncPassword: String): Boolean {
-        val validator = ValidationUtility()
+        val resources = this.resources
+        val validator = ValidationUtility(resources)
         var allCredentialsAreValid = false
 
+        val usernameCredentials = validator.validateUsername(username)
+        val passwordCredentials = validator.validatePasswords(password, vncPassword)
+
         when {
-            username.isEmpty() || password.isEmpty() || vncPassword.isEmpty() -> {
-                Toast.makeText(this, R.string.error_empty_field, Toast.LENGTH_LONG).show()
-            }
-            vncPassword.length > 8 || vncPassword.length < 6 -> {
-                Toast.makeText(this, R.string.error_vnc_password_length_incorrect, Toast.LENGTH_LONG).show()
-            }
-            !validator.isUsernameValid(username) -> {
-                Toast.makeText(this, R.string.error_username_invalid, Toast.LENGTH_LONG).show()
-            }
-            !validator.isPasswordValid(password) -> {
-                Toast.makeText(this, R.string.error_password_invalid, Toast.LENGTH_LONG).show()
-            }
-            !validator.isPasswordValid(vncPassword) -> {
-                Toast.makeText(this, R.string.error_vnc_password_invalid, Toast.LENGTH_LONG).show()
-            }
-            else -> {
+            !usernameCredentials.credentialIsValid ->
+                Toast.makeText(this, usernameCredentials.errorMessage, Toast.LENGTH_LONG).show()
+            !passwordCredentials.credentialIsValid ->
+                Toast.makeText(this, passwordCredentials.errorMessage, Toast.LENGTH_LONG).show()
+            else ->
                 allCredentialsAreValid = true
-                return allCredentialsAreValid
-            }
         }
         return allCredentialsAreValid
     }
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -660,9 +660,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
 
     private fun validateCredentials(username: String, password: String, vncPassword: String): Boolean {
         val blacklistedUsernames = this.resources.getStringArray(R.array.blacklisted_usernames)
-        val validator = ValidationUtility(blacklistedUsernames)
+        val validator = ValidationUtility()
 
-        val usernameCredentials = validator.validateUsername(username)
+        val usernameCredentials = validator.validateUsername(username, blacklistedUsernames)
         val passwordCredentials = validator.validatePassword(password)
         val vncPasswordCredentials = validator.validateVncPassword(vncPassword)
 
--- a/app/src/main/java/tech/ula/utils/ValidationUtility.kt
+++ b/app/src/main/java/tech/ula/utils/ValidationUtility.kt
@@ -1,59 +1,53 @@
 package tech.ula.utils
 
-import android.content.res.Resources
 import tech.ula.R
 import java.util.regex.Matcher
 import java.util.regex.Pattern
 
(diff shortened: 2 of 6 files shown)


app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -1,31 +1,86 @@
...
 
-        val USERNAME_REGEX = "^[a-zA-Z0-9]*\$"
+    fun validateUsername(username: String): Credential {
+        var usernameIsValid = false
+        var errorMessage = ""
+        val blacklist = resources.getStringArray(R.array.blacklisted_usernames)
 
-        pattern = Pattern.compile(USERNAME_REGEX)
-        matcher = pattern.matcher(password)
+        when {
```
MatthewTighe: I think you can do the same trick with the`when` expression here. Something like
```
return when {
    username.isEmpty -> {
        val errorMessage = resources.getString()
        Credential(false, errorMessage)
    }
...
}

File: app/src/main/java/tech/ula/utils/ValidationUtility.kt [hunk 2]
```diff
@@ -1,31 +1,86 @@
...
-        return matcher.matches()
+        return Credential(usernameIsValid, errorMessage)
     }
 
-    fun isPasswordValid(password: String): Boolean {
+    fun validatePasswords(password: String, vncPassword: String): Credential {
+        var passwordIsValid = false
+        var errorMessage = ""
+
+        when {
```
MatthewTighe: Same trick could be employed with this `when`.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -1,31 +1,86 @@
...
+        pattern = Pattern.compile(passwordRegex)
         matcher = pattern.matcher(password)
 
         return matcher.matches()
     }
-}
\ No newline at end of file
+}
+
+sealed class CredentialValidation
```
MatthewTighe: There's no real need for this sealed class here. Data classes can exist without inheriting from a sealed class. I also think the name could be more clear, maybe `CredentialValidationStatus`, or something?

app/src/main/res/values/strings.xml
```diff
@@ -44,6 +44,8 @@
     <string name="error_username_invalid">Username has invalid characters</string>
     <string name="error_password_invalid">Password has invalid characters</string>
     <string name="error_vnc_password_invalid">VNC Password has invalid characters</string>
+    <string name="error_username_invalid_characters">Characters in username must be 1 to 30 characters with lowercase letters and/or numbers.</string>
+    <string name="error_username_in_blacklist">The username: `%1$s` is reserved and cannot be used as a filesystem username.</string>
```
MatthewTighe: Are the backticks surrounding the substitution just for clarity, or do they do something in XML?

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
```diff
@@ -0,0 +1,229 @@
...
+                "uucp",
+                "uuidd",
+                "vboxusers",
+                "video",
+                "voice",
+                "wheel",
+                "www-data"
+        )
+
+        whenever(resources.getStringArray(R.array.blacklisted_usernames))
```
MatthewTighe: You stub the call for `R.array.blacklisted_usernames` twice. I admire the intent to have the whole list here, but again, we're testing _behavior_ and not necessarily exact implementation. We can assume in a real scenario that a call to `resources.getString(R.array.blacklisted_usernames)` will successfully fetch the array, and therefore only need to test it with an array of size 1. This also saves us from needing to keep the lists in sync.

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
```diff
@@ -0,0 +1,229 @@
...
+        whenever(resources.getString(R.string.error_password_invalid))
+                .thenReturn("Password has invalid characters")
+        whenever(resources.getString(R.string.error_vnc_password_invalid))
+                .thenReturn("VNC Password has invalid characters")
+        whenever(resources.getString(R.string.error_vnc_password_length_incorrect))
+                .thenReturn("VNC Password must be between 6 to 8 characters!")
+    }
+
+    @Test
+    fun `Username is empty`() {
```
MatthewTighe: I think these test names could stand to be at least a little more verbose. Something like `Validation fails appropriately if username is empty`.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -1,31 +1,86 @@
...
+data class Credential(val credentialIsValid: Boolean, val errorMessage: String) : CredentialValidation()
```
MatthewTighe: It looks like right now, we're only using the `errorMessage` field to populate toasts. As an FYI, toasts also work with resource ids, which means we could just pass those back. Resource ids are under the app specific `R` package, which means you can import them into the `ValidationUtility`, but you could also access it in the tests, which would save you from needing to assert that the strings are equal (you could just assert the id is the expected). It would also save you from needing to inject a `Resources` object into the utility.

I don't care enough that I think you actually need to change it, because it seems like a fair amount of work for not a ton of benefit, and it's also easy to imagine a scenario where you'd want the error as a string.

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
```diff
@@ -0,0 +1,229 @@
...
+    fun `Username starts with underscore`() {
+        val username = "_123abc"
+        credential = validationUtility.validateUsername(username)
+        assertTrue(credential.credentialIsValid)
+        assertEquals(credential.errorMessage, "")
+    }
+
+    @Test
+    fun `Username is in blacklist`() {
+        val username = "root"
```
MatthewTighe: So for example the array the stub above returns could literally just be `arrayOf("root")`

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -658,31 +658,21 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
         customDialog.show()
     }
 
-    // TODO the view shouldn't be responsible for validation
     private fun validateCredentials(username: String, password: String, vncPassword: String): Boolean {
-        val validator = ValidationUtility()
+        val resources = this.resources
+        val validator = ValidationUtility(resources)
         var allCredentialsAreValid = false
```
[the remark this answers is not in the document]
luongthomas: Yeah I can agree with that.  I've removed all the intermediate booleans such as `allCredentialsAreValid`

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -1,31 +1,86 @@
...
+            else ->
+                usernameIsValid = true
+        }
 
-        return matcher.matches()
+        return Credential(usernameIsValid, errorMessage)
     }
 
-    fun isPasswordValid(password: String): Boolean {
+    fun validatePasswords(password: String, vncPassword: String): Credential {
```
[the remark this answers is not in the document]
luongthomas: I've separated it into two functions and it does make the tests easier to read

app/src/main/java/tech/ula/utils/ValidationUtility.kt
luongthomas: Alright I changed it here too.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
luongthomas: Changed it here too.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
luongthomas: I didn't know that data classes could stand on its own.  I've removed the unnecessary sealed class

app/src/main/res/values/strings.xml
luongthomas: The back ticks were just for clarity but I've removed it in favor of a static string, which makes it easier to test

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
luongthomas: That makes sense, I've changed it to be an array of one.  `root`

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
luongthomas: Updated all test function names

app/src/main/java/tech/ula/utils/ValidationUtility.kt
luongthomas: Good point, I've got it now so that it will compare resource Ids rather than strings, which eliminates the need for injecting the `Resource` object

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -1,59 +1,53 @@
 package tech.ula.utils
 
-import android.content.res.Resources
 import tech.ula.R
 import java.util.regex.Matcher
 import java.util.regex.Pattern
 
-class ValidationUtility(systemResources: Resources) {
+class ValidationUtility(blacklisted_usernames: Array<String>) {
```
MatthewTighe: Even though `blacklist_usernames` is being retrieved from XML which uses underscores, the parameter name should still probably be camelcase to follow Kotlin convention. It also might make more sense to send the list as a function parameter to `validateUsername` rather than as a class dependency for the `ValidationUtility`.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -82,5 +76,4 @@ class ValidationUtility(systemResources: Resources) {
     }
 }
 
-sealed class CredentialValidation
-data class Credential(val credentialIsValid: Boolean, val errorMessage: String) : CredentialValidation()
\ No newline at end of file
+data class Credential(val credentialIsValid: Boolean, val errorMessageId: Int = 0)
```
MatthewTighe: I still don't love this class name, as I don't think `Credential` really encapsulates the concept of the data it holds.

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
```diff
@@ -24,206 +23,129 @@ class ValidationUtilityTest {
...
     }
 
     @Test
-    fun `vnc password is empty`() {
-        val password = defaultPassword
+    fun `Validation fails appropriately if vnc password is empty`() {
         val vncPassword = ""
 
-        credential = validationUtility.validatePasswords(password, vncPassword)
+        credential = validationUtility.validatePassword(vncPassword)
```
MatthewTighe: You call `validatePassword` instead of `validateVncPassword` here.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
luongthomas: Changed now to CredentialValidationStatus

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
luongthomas: Good catch!

PR #608, opened 2019-02-10
Improve Chinese language pack

SunbossRS:
I am a native Chinese and i use the UserLAnd to find some language bug.
Some language string is not full.

### DIFF
--- a/app/src/main/res/values-zh-rCN/strings.xml
+++ b/app/src/main/res/values-zh-rCN/strings.xml
@@ -106,6 +106,6 @@
         点击“是的”将会清除支持的文件。但这不会发生在安装的应用程序的文件的。
         点击“是的”将会导致在下次运行会话或应用程序时再下载资源。
     </string>
-    <string name="alert_clear_support_files_clear_button">是的</string>
+    <string name="alert_clear_support_files_clear_button">清除</string>
 
 </resources>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -75,24 +81,31 @@
...
     <string name="github_message">发现错误并且想要申请新的功能？上我们的Github，在那里您能得到最快的答复！</string>
     <string name="pref_proot_debugging_level_summary">数值越大，越冗余越高.</string>
-</resources>
\ No newline at end of file
+    <string name="alert_clear_support_files_title">真的要清除支持文件吗?</string>
+    <string name="alert_clear_support_files_message">
+        点击“是的”将会清除支持的文件。不过不要担心你的应用程序，我们是不会清除应用程序的文件的。
+        点击“是的”将会导致在下次运行会话或应用程序时再下载资源。
+    </string>
+    <string name="alert_clear_support_files_clear_button">是的</string>
```
AndrewScibek: This means yes not clear. @MatthewTighe where does this get used?

AndrewScibek:
Other than my one comment everything looks good

app/src/main/res/values-zh-rCN/strings.xml
SunbossRS: But in Chinese version UserLAnd there isn't this thing

AndrewScibek: well i was thinking 清除 or 删除 might make a little more sense

SunbossRS: 清除 can say "clear",and 删除 can say "delete"

MatthewTighe: This is used in the options menu that can be opened in the top-right of the screen. It is meant for deleting all the support files we download.

AndrewScibek: @SunbossRS can you change it to say 清除 and then we can pull in the translations

AndrewScibek:
We need to build on forked prs.

PR #615, opened 2019-02-11
Handle syncing download state onResume.

MatthewTighe:
**Describe the pull request**

We previously were only syncing the download state when the session fsm was being initialized. This solved the issue of process death, but did not solve the issue of the app being put into the background but not being killed. Now, the download state will be synced whenever the lifecycle method `onResume` is called on the main activity.

**Link to relevant issues**
N/A


EnderNightLord-ChromeBook:
How Does It Work?

MatthewTighe:
By executing the cache syncing flow from `MainActivity#onResume`

EnderNightLord-ChromeBook:
Ohhh That Makes Sense... Just Re Read It.

PR #628, opened 2019-02-12
User review request

luongthomas:
**Describe the pull request**

Provides a way to ask the user for feedback/play store rating after it appears the user has been using the app frequently.  

Will show an unobtrusive UI element to request feedback which is was inspired by [a Medium post](https://medium.com/circa/the-right-way-to-ask-users-to-review-your-app-9a32fd604fca) by Circa.

The UI element will only appear if the user has opened the app at least 15 times and it has been 3 days since the first time they have opened it.  Once the user has either given feedback or dismissed the UI element, it will not appear again.  This process will reset if the app data is cleared.

## If the user wants to give a rating, it will direct them to the Play Store.

**Link to relevant issues**
N/A

### DIFF
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -80,8 +80,8 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
         NotificationUtility(this)
     }
 
-    private val usageUtility by lazy {
-        UsageUtility(this.getSharedPreferences("usage", Context.MODE_PRIVATE))
+    private val userFeedbackPrefs by lazy {
+        UserFeedbackPreferences(this.getSharedPreferences("usage", Context.MODE_PRIVATE))
     }
 
     private val downloadBroadcastReceiver = object : BroadcastReceiver() {
@@ -153,7 +153,7 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
 
         setupWithNavController(bottom_nav_view, navController)
 
-        usageUtility.addToNumberOfTimesOpened()
+        userFeedbackPrefs.addToNumberOfTimesOpened()
         setupReviewRequest()
 
         viewModel.getState().observe(this, stateObserver)
@@ -171,9 +171,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
     }
 
     private fun setupReviewRequest() {
-        val minimumTimePassedReached = usageUtility.getIsSufficientTimeElapsedSinceFirstOpen()
-        val minimumTimesOpenedReached = usageUtility.numberOfTimesOpenedIsGreaterThanThreshold()
-        val userAlreadyGaveReviewOrDismissed = usageUtility.getUserGaveFeedback()
+        val minimumTimePassedReached = userFeedbackPrefs.getIsSufficientTimeElapsedSinceFirstOpen()
+        val minimumTimesOpenedReached = userFeedbackPrefs.numberOfTimesOpenedIsGreaterThanThreshold()
+        val userAlreadyGaveReviewOrDismissed = userFeedbackPrefs.getUserGaveFeedback()
 
         if (minimumTimePassedReached && minimumTimesOpenedReached && !userAlreadyGaveReviewOrDismissed) {
             val requestReviewView = layoutInflater.inflate(R.layout.list_item_review_request, null)
@@ -225,7 +225,7 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
     }
 
     private fun userGaveReviewOrDismissed(viewHolder: ViewGroup) {
-        usageUtility.setUserGaveFeedback()
+        userFeedbackPrefs.setUserGaveFeedback()
         viewHolder.visibility = View.GONE
     }
 
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -175,7 +175,7 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
         val minimumTimesOpenedReached = usageUtility.numberOfTimesOpenedIsGreaterThanThreshold()
         val userAlreadyGaveReviewOrDismissed = usageUtility.getUserGaveFeedback()
 
-        if (minimumTimePassedReached && minimumTimesOpenedReached && userAlreadyGaveReviewOrDismissed) {
+        if (minimumTimePassedReached && minimumTimesOpenedReached && !userAlreadyGaveReviewOrDismissed) {
             val requestReviewView = layoutInflater.inflate(R.layout.list_item_review_request, null)
             requestReviewView.visibility = View.VISIBLE
             val viewHolder = findViewById<ViewGroup>(R.id.request_review_insert_point)
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -25,7 +25,9 @@ import android.support.v7.app.AppCompatActivity
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
+import android.view.ViewGroup
 import android.view.animation.AlphaAnimation
+import android.widget.Button
 import android.widget.RadioButton
 import android.widget.TextView
 import android.widget.Toast
@@ -76,6 +78,10 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
 
     private val notificationManager by lazy {
         NotificationUtility(this)
+    }
+
+    private val usageUtility by lazy {
+        UsageUtility(this.getSharedPreferences("usage", Context.MODE_PRIVATE))
     }
 
     private val downloadBroadcastReceiver = object : BroadcastReceiver() {
@@ -147,6 +153,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
 
         setupWithNavController(bottom_nav_view, navController)
 
+        usageUtility.addToNumberOfTimesOpened()
+        setupReviewRequest()
+
         viewModel.getState().observe(this, stateObserver)
     }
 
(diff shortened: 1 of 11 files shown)


app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -398,4 +399,54 @@ class AcraWrapper {
...
+                putInt(numberOfTimesOpenedKey, numberTimesOpened + 1)
+                apply()
+            }
+        }
+
+        return numberTimesOpened
+    }
+
+    fun getIsSufficientTimeElapsedSinceFirstOpen(): Boolean {
+        val dateTimeFirstOpened = prefs.getLong(dateTimeFirstOpenKey, 0L)
```
MatthewTighe: I think you could combine `getIsSufficientTimeElapsedSinceFirstOpen`, `numberOfTimesOpenedIsGreaterThanThreshold`, and `getUserGaveFeedback` into one method, e.g. `UsageUtility#askingForFeedbackIsAppriopriate`.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -398,4 +399,54 @@ class AcraWrapper {
     fun putCustomString(key: String, value: String) {
         ACRA.getErrorReporter().putCustomData(key, value)
     }
+}
+
+class UsageUtility(private val prefs: SharedPreferences) {
```
MatthewTighe: I think this class name could be more descriptive, maybe something like `UserFeedbackUtility` or `UserFeedbackPreferences` to match other preference wrappers.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -398,4 +399,54 @@ class AcraWrapper {
...
+    private val millisecondsInThreeDays = 259200000L
+    private val minimumNumberOfOpensBeforeReviewRequest = 15
+
+    fun addToNumberOfTimesOpened(): Int {
+        val numberTimesOpened = prefs.getInt(numberOfTimesOpenedKey, 1)
+
+        if (numberTimesOpened == 1) {
+            with(prefs.edit()) {
+                putInt(numberOfTimesOpenedKey, numberTimesOpened + 1)
+                putLong(dateTimeFirstOpenKey, Date().time)
```
MatthewTighe: You can use `System.currentTimeMillis` to avoid potential issues with `Date` and more closely match what we do elsewhere.

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -161,6 +170,65 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
         ACRA.init(applicationContext as Application, builder)
     }
 
+    private fun setupReviewRequest() {
+        val minimumTimePassedReached = usageUtility.getIsSufficientTimeElapsedSinceFirstOpen()
+        val minimumTimesOpenedReached = usageUtility.numberOfTimesOpenedIsGreaterThanThreshold()
+        val userAlreadyGaveReviewOrDismissed = usageUtility.getUserGaveFeedback()
+
+        if (minimumTimePassedReached && minimumTimesOpenedReached && !userAlreadyGaveReviewOrDismissed) {
```
MatthewTighe: If you follow the suggestion below, you could change this to simply 
`if (feedbackUtility.askingForFeedbackIsAppropriate)  setupReviewRequestUI()`
and then you can move the rest of the UI setup there as well.

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -161,6 +170,65 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
...
+                startActivity(intent)
+            }
+
+            negativeBtn.setOnClickListener {
+                userGaveReviewOrDismissed(viewHolder)
+            }
+        }
+    }
+
+    private fun userGaveReviewOrDismissed(viewHolder: ViewGroup) {
```
MatthewTighe: I think this function could more appropriately be named `handleUserFeedbackComplete` or something like that.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -398,4 +399,54 @@ class AcraWrapper {
...
+}
+
+class UsageUtility(private val prefs: SharedPreferences) {
+    private val numberOfTimesOpenedKey = "numberOfTimesOpened"
+    private val userGaveFeedbackKey = "userGaveFeedback"
+    private val dateTimeFirstOpenKey = "dateTimeFirstOpen"
+    private val millisecondsInThreeDays = 259200000L
+    private val minimumNumberOfOpensBeforeReviewRequest = 15
+
+    fun addToNumberOfTimesOpened(): Int {
```
MatthewTighe: I like `incrementNumberOfTimesOpened` more personally, but that's nitpicky.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -398,4 +399,54 @@ class AcraWrapper {
...
+        val numberTimesOpened = prefs.getInt(numberOfTimesOpenedKey, 1)
+
+        if (numberTimesOpened == 1) {
```
MatthewTighe: I think you could move the the `if` statement into `with` to look a bit cleaner.

```
with (prefs.edit()) {
    putInt(numberOfTimesOpenedKey, numberTimesOpened + 1)
    if (numberTimesOpened == 1) putLong(firstTimeOpenedMillis, System.currentTimeMillis)
    apply()
}
```

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -398,4 +399,54 @@ class AcraWrapper {
...
+
+        return (Date().time > dateTimeWithSufficientTimeElapsed)
+    }
+
+    fun numberOfTimesOpenedIsGreaterThanThreshold(): Boolean {
+        val numberTimesOpened = prefs.getInt(numberOfTimesOpenedKey, 1)
+        return numberTimesOpened > minimumNumberOfOpensBeforeReviewRequest
+    }
+
+    fun setUserGaveFeedback() {
```
MatthewTighe: What do you think about `UserHasGivenFeedback` instead? Reads a bit cleaner to me

app/src/main/res/values/strings.xml
```diff
@@ -135,6 +135,10 @@
     <string name="client_xsdl">XSDL</string>
 
     <!-- Alerts -->
+    <string name="button_yes">Yes!</string>
+    <string name="button_negative_1">Not really</string>
+    <string name="button_negative_2">No, thanks</string>
```
MatthewTighe: Could you change the names of these buttons to be more descriptive than numbers?

app/src/main/res/layout/activity_main.xml
```diff
@@ -1,36 +1,37 @@
 <?xml version="1.0" encoding="utf-8"?>
-<android.support.constraint.ConstraintLayout
+
+<LinearLayout
```
MatthewTighe: Using linear layouts seems to break rotation on my device. I get a 4th soft input button that I have to click to rotate the screen.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I made a function called `askingForFeedbackIsAppropriate` while hiding the other three as private functions.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I changed it to `UserFeedbackUtility` to match the other wrappers

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I've changed it to use `System.currentTimeMillis` instead of using `Date`

app/src/main/java/tech/ula/MainActivity.kt
luongthomas: I've changed it to `handleUserFeedback`

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I agree with that name 👍 I've changed it to `incrementNumberOfTimesOpened`

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: Oh yeah that looks a lot cleaner, I've changed it to move the if statement into the `with` block

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: That's a great name, I've changed it to `UserHasGivenFeedback`

app/src/main/res/layout/activity_main.xml
luongthomas: I've reverted it back to a constraint layout.  Let me know if that improves the rotations on your device

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -401,52 +398,50 @@ class AcraWrapper {
...
-                apply()
-            }
-        }
+    fun askingForFeedbackIsAppropriate(): Boolean {
+        val minimumTimePassedReached = getIsSufficientTimeElapsedSinceFirstOpen()
+        val minimumTimesOpenedReached = numberOfTimesOpenedIsGreaterThanThreshold()
+        val userAlreadyGaveReviewOrDismissed = getUserGaveFeedback()
 
-        return numberTimesOpened
+        return minimumTimePassedReached && minimumTimesOpenedReached && !userAlreadyGaveReviewOrDismissed
```
MatthewTighe: This might be handled by the compiler, but you could save a bit of work by calling the functions inline in the return. That way, as soon as one fails, the other functions won't be even be called.

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    private fun appOpenedNumberOfDaysAgo(days: Int) {
+        val millisecondsInOneDay = 86400000L
+        val millisecondsNow = System.currentTimeMillis()
+        whenever(sharedPrefs.getLong("dateTimeFirstOpen", 0))
+                .thenReturn(millisecondsNow - (millisecondsInOneDay * days))
+    }
+
+    private fun setTimesAppOpened(numberOfTimesOpened: Int) {
+        whenever(sharedPrefs.getInt("numberOfTimesOpened", 1))
```
MatthewTighe: Wrapping stubs like this is genius. Definitely stealing this method.

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun `Asking for feedback is appropriate if app opened if app opened four days ago`() {
+        setTimesAppOpened(minimumTimesAppOpenedToShowReview)
+        setUserGaveFeedback(false)
+        appOpenedNumberOfDaysAgo(4)
+
+        assertTrue(userFeedbackUtility.askingForFeedbackIsAppropriate())
+    }
+
+    @Test
+    fun `Asking for feedback is not appropriate if app opened if app opened two days ago`() {
```
MatthewTighe: Some repeated text in this function name.

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    @Test
+    fun `Asking for feedback is not appropriate when app wasn't opened more than the required amount`() {
+        setTimesAppOpened(minimumTimesAppOpenedToShowReview - 5)
+        setUserGaveFeedback(false)
+        appOpenedNumberOfDaysAgo(3)
+
+        assertFalse(userFeedbackUtility.askingForFeedbackIsAppropriate())
+    }
+
+    private fun appOpenedNumberOfDaysAgo(days: Int) {
```
MatthewTighe: I've been trying to structure test files such that they go:

mocks
global test variables
setup functions (including `@Before`) but also stuff like these private functions.

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -170,62 +171,54 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
...
-        if (minimumTimePassedReached && minimumTimesOpenedReached && !userAlreadyGaveReviewOrDismissed) {
-            val requestReviewView = layoutInflater.inflate(R.layout.list_item_review_request, null)
-            requestReviewView.visibility = View.VISIBLE
-            val viewHolder = findViewById<ViewGroup>(R.id.request_review_insert_point)
-            viewHolder.addView(requestReviewView, 0)
-            setupReviewRequestUI(viewHolder)
-        }
-    }
+        requestReviewView.visibility = View.VISIBLE
+        viewHolder.addView(requestReviewView, 0)
```
MatthewTighe: I'll preface this by saying that views, view groups, layout inflation etc. is still one of the Android areas I feel weakest on. It looks like this could be changed to potentially be more performant, and avoid the IDE warning that using `null` in the call to `inflate` gives.

If you remove the `android:visibility` attribute from `list_item_review_request`, initialize the `viewHolder` first, you can then use the `viewHolder` as the `root` parameter to the `inflate` call. You'd then be able to remove the following two lines that set visibility and add the view the the `viewHolder`. Then in `handleUserFeedback`, you can call `viewHolder.removeAllViews()` instead of adjusting the visibility. I think this should reduce memory footprint, and if I remember right views are one of the more costly pieces of Android.

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.junit.MockitoJUnitRunner
+
+@RunWith(MockitoJUnitRunner::class)
+class UserFeedbackUtilityTest {
```
MatthewTighe: Don't forget to test the successful case as well.

app/src/main/java/tech/ula/MainActivity.kt
luongthomas: That's a good idea.  I did notice an improvement in performance when navigating to different fragments when switching to this approach

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
luongthomas: I've added a successful test case

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
luongthomas: Moved the private setup functions above the test functions

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
luongthomas: Good catch, I've removed the extra words

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I've changed the function so it calls all three sub-functions inline

app/src/test/java/tech/ula/utils/UserFeedbackUtilityTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    fun `Asking for feedback is appropriate if app opened four days ago, hasn't given feedback and opened the app at the minimum amount of times`() {
+        setTimesAppOpened(minimumTimesAppOpenedToShowReview)
+        setUserGaveFeedback(false)
+        appOpenedNumberOfDaysAgo(4)
+
+        assertTrue(userFeedbackUtility.askingForFeedbackIsAppropriate())
+    }
+
+    @Test
+    fun `Asking for feedback is appropriate if app opened four days ago`() {
```
MatthewTighe: It looks like this test is repeated. Sorry if I just missed it the first time 😅

luongthomas: Removed 👍

Issue #614, opened 2019-02-11
Acquire wake lock crash


XeGrox:
Acquire wake lock action from the terminal notification causes the app including the running service to crash after updating to the latest version (2.3.0 to 2.3.1)

Emui 9

MatthewTighe:
I can confirm this is reproducible. Not sure what would've changed in that version to break it, but we'll take a look. @luongthomas can you take this one?

luongthomas:
I'll investigate what's causing this crash

lewisdonofrio:
I've been seening these as well even in most recent 2.3.4

slycooper724:
It happened to me too.

lewisdonofrio:
same with 2.3.7 ;(

luongthomas:
Release v2.3.8 is being made to production to fix this issue momentarily.

lewisdonofrio:
it fixed the issue...tnx

PR #663, opened 2019-02-27
Fix crash related to attempting to acquire a wakelock

luongthomas:
**Describe the pull request**

This PR addresses the issue when a user wants to acquire a wakelock through the UserLAnd terminal notification bar.  The error message that was displayed upon crash was that the app did not have permissions for using wakelock.  

The solution to this problem can be found in this [StackOverflow post](https://stackoverflow.com/questions/50972669/cannot-get-wake-lock-permission-on-android-8-oreo/50972790#50972790).

Including this PR change will enable wakelock permissions for the terminal again and prevent the crash from occurring.  

debugging is difficult

**Link to relevant issues**
 #614

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -172,7 +172,7 @@ task jacocoCoverageReportForCi(type: JacocoReport, dependsOn: ['testDebugUnitTes
 }
 
 dependencies {
-    def acra_version = '5.1.3'
+    def acra_version = '5.2.1'
 
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':terminal-term')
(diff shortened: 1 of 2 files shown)


luongthomas:
> Wow, good find. I think it would be better to just update our ACRA version to the latest release candidate rather than implementing this workaround. We should try to keep our library versions up to date anyway, and it'll be easier to keep track of version updates than remembering to remove the workaround when it becomes unnecessary.
> 
> Really wish these merged manifests were more transparent. This is the second time an issue like this has messed with our build.
> 
> Thanks for digging this one up!

Upgrading the ACRA version fixes the issue as well, thanks for pointing it out.  I've made changes to the PR accordingly 😄

PR #664, opened 2019-02-28
Fix following symlinks when deleting filesystems

### LINKED ISSUE #275: Filesystems->delete follows symlinks into /sdcard/
Deleting a filesystem in the UserLAnd GUI deletes that filesystem and recursively follows symbolic links within the filesystem into the /sdcard/ directory; files underneath linked directories are deleted. I just deleted almost everything on my mobile...

- create a directory 'test' in your internal storage, create some files underneath.
- create a filesystem 'debian' in UserLAnd
- log into 'debian' and 'ln -s /sdcard/test ./test; exit'
- delete the 'debian' filesystem in UserLAnd
- ... and find 'test' empty.

I did expect only 'debian' to be removed but my files in /sdcard/, wich are shared among differend UserLAnd filesystems, to remain untouched.

Device: Moto X4
Android Version: 8.0.0
UserLAnd Version 0.3.4

MatthewTighe:
**Describe the pull request**

This fixes the issue in which deleting a filesystem would following symlinks that had been created to areas outside of the filesystem. Previously, we had been using Kotlin's `File.deleteRecursively()` extension function to delete filesystems, but this will follow symlinks. Now, we use an a newly refactored `BusyboxExecutor` (old version was `ExecUtility`) to execute `rm -rf` as a busybox command.

Good places to start could be the new `BusyboxExecutor` class or the `Delete Filesystem` action taken from the `FilesystemListFragment`

**Link to relevant issues**
Closes #275


app/src/main/java/tech/ula/utils/FilesystemUtility.kt
```diff
@@ -81,19 +84,12 @@ class FilesystemUtility(
...
-                directory.deleteRecursively()
-            val isDirectoryDeleted = directory.deleteRecursively()
-            if (isDirectoryDeleted) {
-                val successMessage = "Successfully deleted filesystem located at: $directory"
-                logger.v("Filesystem Delete", successMessage)
-            } else {
-                val errorMessage = "Error in attempting to delete filesystem located at: $directory"
-                logger.e("Filesystem Delete", errorMessage)
-            }
+        if (!directory.exists() || !directory.isDirectory) return
```
luongthomas: Can we change the `directory` variable to be called `filesystem` instead, it would make it easier to read.
For example: If the filesystem doesn't exist or the filesystem isn't a directory, then return.

app/src/main/java/tech/ula/utils/ServerUtility.kt
```diff
@@ -51,75 +55,90 @@ class ServerUtility(
...
+        // The third-party app is responsible for handling errors starting their server.
+        if (session.serviceType == "xsdl") return true
+        val result = busyboxExecutor.executeCommand(command)
+        return when (result) {
+            is SuccessfulExecution -> true
+            is FailedExecution -> {
+                logger.logRuntimeErrorForCommand(functionName = "isServerRunning", command = command, err = result.reason)
+                false
+            }
+            else -> {
```
luongthomas: The brackets holding `false` could be removed here.

app/src/test/java/tech/ula/utils/AssetFileClearerTest.kt
```diff
@@ -72,25 +78,27 @@ class AssetFileClearerTest {
...
     @Test
-    fun `Clears all assets and leaves filesystem structure intact`() {
+    fun `Clears all assets and leaves filesystem structure intact`() = runBlocking {
+        whenever(busyboxExecutor.recursivelyDelete(any())).thenReturn(SuccessfulExecution)
         assetFileClearer.clearAllSupportAssets()
 
-        assertTrue(filesDir.exists())
-        assertTrue(randomTopLevelFile.exists())
-        assertTrue(randomTopLevelDir.exists())
+        verify(busyboxExecutor, never()).recursivelyDelete(filesDir.absolutePath)
```
luongthomas: That's a really neat feature of Mockito!  I'll keep this in mind when writing unit tests

MatthewTighe:
Alright, I've updated according to your review.

PR #675, opened 2019-03-05
Compression implementation

MatthewTighe:
**Describe the pull request**

Add implementation for compression and some tests.

**Link to relevant issues**
N/A

### DIFF
--- a/app/src/main/java/tech/ula/utils/FilesystemUtility.kt
+++ b/app/src/main/java/tech/ula/utils/FilesystemUtility.kt
@@ -61,8 +61,26 @@ class FilesystemUtility(
         }
     }
 
-    suspend fun compressFilesystem(filesystem: Filesystem, externalStorageDirectory: File) = withContext(Dispatchers.IO) {
-        Unit
+    suspend fun compressFilesystem(filesystem: Filesystem, externalStorageDirectory: File, listener: (String) -> Any) = withContext(Dispatchers.IO) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/tech/ula/utils/FilesystemUtility.kt
```diff
@@ -61,8 +61,26 @@ class FilesystemUtility(
...
     }
 
-    suspend fun compressFilesystem(filesystem: Filesystem, externalStorageDirectory: File) = withContext(Dispatchers.IO) {
-        Unit
+    suspend fun compressFilesystem(filesystem: Filesystem, externalStorageDirectory: File, listener: (String) -> Any) = withContext(Dispatchers.IO) {
+        val filesystemDirName = "${filesystem.id}"
+        val command = "/support/common/compressFilesystem.sh"
+
+        val externalUserlandDirPath = "${externalStorageDirectory.absolutePath}/UserLAnd"
+        val backupName = "${filesystem.name}-${filesystem.distributionType}-rootfs.tar.gz"
```
luongthomas: We'll have to address the possible issue of invalid filesystem name characters being used in a filename later on

MatthewTighe: That's a good point. Think we can just update the `ValidationUtility` to also validate filesystem names, and do so during creation. I'll put a card up, because think that needs to get done before this feature is complete.

PR #676, opened 2019-03-06
Add UI for filesystem backups import and export

luongthomas:
**Describe the pull request**

This will allow users to navigate to the filesystemListFragment and longpress an existing filesystem in order to export their filesystem off of UserLAnd.  

It will be stored in their external memory.   In the Kotlin code it will be saved via:
`Environment.getExternalStoragePublicDirectory("UserLAnd")`

Importing the same file to use it as a filesystem is also supported too, when in the FilesystemEditFragment menu, there will be a "Show Advanced Options" button.  

Screen Shot 2019-03-12 at 12 06 34 PM

Upon pressing it, the import button will be visible along with a description of what the user should fill out in the text fields above.

Screen Shot 2019-03-12 at 12 07 02 PM

**Link to relevant issues**

### DIFF
--- a/app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
+++ b/app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
@@ -2,8 +2,11 @@ package tech.ula.ui
 
 import android.app.Activity
 import android.arch.lifecycle.ViewModelProviders
+import android.content.ActivityNotFoundException
 import android.content.Context
+import android.content.Intent
 import android.os.Bundle
+import android.support.design.widget.Snackbar
 import android.support.v4.app.Fragment
 import android.text.Editable
 import android.text.TextWatcher
@@ -89,6 +92,7 @@ class FilesystemEditFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
 
         setupTextInputs()
+        setupImportButton()
 
         if (editExisting) {
             spinner_filesystem_type.isEnabled = false
@@ -153,6 +157,31 @@ class FilesystemEditFragment : Fragment() {
         })
     }
 
+    private fun setupImportButton() {
+        import_button.setOnClickListener {
+            val filePickerIntent = Intent(Intent.ACTION_GET_CONTENT)
+            filePickerIntent.type = "application/*"
+            filePickerIntent.addCategory(Intent.CATEGORY_OPENABLE)
+            val fileChooser = Intent.createChooser(filePickerIntent, "Select File")
+
+            try {
+                startActivityForResult(fileChooser, IMPORT_FILESYSTEM_REQUEST_CODE)
+            } catch (activityNotFoundErr: ActivityNotFoundException) {
+                Toast.makeText(activityContext, "Please install a File Manager.", Toast.LENGTH_LONG).show()
+            }
+        }
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(requestCode, resultCode, data)
+        if (requestCode == IMPORT_FILESYSTEM_REQUEST_CODE) {
+            data?.data?.let {
+                val path = it.path
+                Toast.makeText(activityContext, "Location is: $path", Snackbar.LENGTH_LONG).show()
+            }
+        }
+    }
+
     private fun insertFilesystem(): Boolean {
         val navController = NavHostFragment.findNavController(this)
         if (!filesystemParametersAreCorrect()) {
@@ -203,5 +232,9 @@ class FilesystemEditFragment : Fragment() {
                 return true
         }
         return false
+    }
+
+    companion object {
+        val IMPORT_FILESYSTEM_REQUEST_CODE = 5
     }
 }
--- a/app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
+++ b/app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
@@ -1,10 +1,14 @@
 package tech.ula.ui
 
+import android.Manifest
+import android.annotation.TargetApi
+import android.app.AlertDialog
+import android.arch.lifecycle.Observer
 import android.arch.lifecycle.ViewModelProviders
 import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
+import android.os.Build
 import android.os.Bundle
 import android.support.v4.app.Fragment
 import android.text.Editable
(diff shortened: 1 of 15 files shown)


app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -39,6 +50,19 @@ class FilesystemEditFragment : Fragment() {
...
     }
 
+    private val permissionRequestCode: Int by lazy {
+        getString(R.string.permission_request_code).toInt()
+    }
+
+    private val filesystemImportStatusObserver = Observer<FilesystemImportStatus> {
+        it?.let { importStatus ->
+            when (importStatus) {
+                is ImportSuccess -> Toast.makeText(activityContext, "Successfully imported backup", Toast.LENGTH_LONG).show()
```
MatthewTighe: I think these should be string resources.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -153,6 +180,72 @@ class FilesystemEditFragment : Fragment() {
         })
     }
 
+    private fun setupImportButton() {
+        import_button.setOnClickListener {
+            val filePickerIntent = Intent(Intent.ACTION_GET_CONTENT)
+            filePickerIntent.addCategory(Intent.CATEGORY_OPENABLE)
+            filePickerIntent.type = "application/*"
+            val fileChooser = Intent.createChooser(filePickerIntent, "Select Filesystem Backup file")
```
MatthewTighe: If `Select Filesystem Backup file` can be a string resource, I think it should be. Capitalization could also be made uniform.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -153,6 +180,72 @@ class FilesystemEditFragment : Fragment() {
...
+            if (!arePermissionsGranted(activityContext)) {
+                showPermissionsNecessaryDialog()
+                return@setOnClickListener
+            }
+
+            try {
+                filesystem.isCreatedFromBackup = true
+                startActivityForResult(fileChooser, IMPORT_FILESYSTEM_REQUEST_CODE)
+            } catch (activityNotFoundErr: ActivityNotFoundException) {
+                Toast.makeText(activityContext, "Please install a File Manager.", Toast.LENGTH_LONG).show()
```
MatthewTighe: I think this should be a string resource.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -204,4 +301,8 @@ class FilesystemEditFragment : Fragment() {
         }
         return false
     }
+
+    companion object {
```
MatthewTighe: I think this companion object can just be a private val.

app/src/main/java/tech/ula/ui/FilesystemListFragment.kt
```diff
@@ -46,6 +51,24 @@ class FilesystemListFragment : Fragment() {
...
     }
 
+    private val filesystemExportStatusObserver = Observer<FilesystemExportStatus> {
+        it?.let { exportStatus ->
+            when (exportStatus) {
+                is ExportUpdate -> {
+                    activityContext.updateExportProgress(exportStatus.details)
+                }
+                is ExportSuccess -> {
+                    Toast.makeText(activityContext, "Successfully exported backup", Toast.LENGTH_LONG).show()
```
MatthewTighe: I think these toasts should be string resources.

luongthomas:
The tests are fixed now and I've made all the string literals into a string resource

app/src/main/java/tech/ula/ui/FilesystemListFragment.kt
```diff
@@ -46,6 +51,27 @@ class FilesystemListFragment : Fragment() {
...
+            when (exportStatus) {
+                is ExportUpdate -> {
+                    activityContext.updateExportProgress(exportStatus.details)
+                }
+                is ExportSuccess -> {
+                    Toast.makeText(activityContext, R.string.export_success, Toast.LENGTH_LONG).show()
+                    activityContext.stopExportProgress()
+                }
+                is ExportFailure -> {
+                    Toast.makeText(activityContext, "${R.string.export_failure} + ${exportStatus.reason}", Toast.LENGTH_LONG).show()
```
MatthewTighe: Does `${R.string.export_failure} + ${exportStatus.reason}` work as expected? Seems to me like it would like just call `toString` on the `Int` itself. This could be fixed by either using string formatting in the string resource itself, or by calling `getString` on the resource and prepending the result to `exportStatus.reason`.

app/src/main/java/tech/ula/viewmodel/FilesystemEditViewModel.kt
```diff
@@ -18,28 +29,68 @@ class FilesystemEditViewModel(private val ulaDatabase: UlaDatabase) : ViewModel(
...
+        return importStatusLiveData
+    }
+
     fun insertFilesystem(filesystem: Filesystem, coroutineScope: CoroutineScope = this) = coroutineScope.launch {
         withContext(Dispatchers.IO) {
             ulaDatabase.filesystemDao().insertFilesystem(filesystem)
         }
     }
 
+    fun getBackupUri(): Uri? {
```
MatthewTighe: I think since we're exposing both a getter and setter for this field that we should just make it public.

app/src/main/java/tech/ula/ui/FilesystemListFragment.kt
MatthewTighe: Additionally, I think failure modes could be better served by a dialog than a toast. That way, the user will have a better chance to understand the failure as well as report it to us.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -39,6 +52,19 @@ class FilesystemEditFragment : Fragment() {
...
 
+    private val permissionRequestCode: Int by lazy {
+        getString(R.string.permission_request_code).toInt()
+    }
+
+    private val filesystemImportStatusObserver = Observer<FilesystemImportStatus> {
+        it?.let { importStatus ->
+            when (importStatus) {
+                is ImportSuccess -> Toast.makeText(activityContext, R.string.import_success, Toast.LENGTH_LONG).show()
+                is ImportFailure -> Toast.makeText(activityContext, R.string.import_failure, Toast.LENGTH_LONG).show()
```
MatthewTighe: Same thing about a dialog here.

PR #678, opened 2019-03-07
Fix problem formatting text with images

foodmaher:
Fix problem formatting text with images for different languages

There are errors in formatting the text in some languages ​​written form the opposite side for English where the text appears on the image I have corrected this problem according to my experience modest

LinearLayout format is used because it prevents the appearance of elements above some and also determines the horizontal width instead of the vertical

Please review this edit on Android studio and make sure that it works as required in all languages. Or try to correct this problem by he is references

**The AIDE application was used**
screenshot_2019-03-07-11-49-37


corbinlc:
Why was this closed?  Do you think it was the wrong solution?

PR #701, opened 2019-03-14
Do not allow exporting backup while session is active

luongthomas:
**Describe the pull request**

When a user attempts to export a filesystem while a session is active, tell them to turn off any active sessions before continuing.

ezgif com-video-to-gif(5)

**Link to relevant issues**
N/A

### DIFF
--- a/app/src/main/java/tech/ula/viewmodel/FilesystemListViewModel.kt
+++ b/app/src/main/java/tech/ula/viewmodel/FilesystemListViewModel.kt
@@ -4,7 +4,9 @@ import android.arch.lifecycle.LiveData
 import android.arch.lifecycle.MutableLiveData
 import android.arch.lifecycle.ViewModel
 import android.arch.lifecycle.ViewModelProvider
+import android.os.Environment
 import kotlinx.coroutines.* // ktlint-disable no-wildcard-imports
+import tech.ula.R
 import tech.ula.model.daos.FilesystemDao
 import tech.ula.model.daos.SessionDao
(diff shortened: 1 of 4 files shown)


luongthomas:
Moved the logic to the ViewModel and fixed some tests

app/src/main/java/tech/ula/viewmodel/FilesystemListViewModel.kt
```diff
@@ -78,25 +80,34 @@ class FilesystemListViewModel(private val filesystemDao: FilesystemDao, private
...
                 true -> exportStatusLiveData.postValue(ExportSuccess)
-                false -> exportStatusLiveData.postValue(ExportFailure("Exporting to external directory failed, exported file has no data"))
+                false -> exportStatusLiveData.postValue(ExportFailure(R.string.error_export_to_external_failed_no_data))
             }
         }
     }
+
+    fun startExport(filesystem: Filesystem, activeSessions: List<Session>, filesDir: File) {
+        if (activeSessions.isEmpty()) {
+            val externalDestination = Environment.getExternalStoragePublicDirectory("UserLAnd")
```
MatthewTighe: By using `Environment` at this level, I don't think this function will be unit-testable. I think it's best to pass the `externalDestination` down as a parameter, so that we can use test files and folders as necessary.

Could you also write a test to make sure the behavior of only allowing exports while sessions aren't active doesn't regress?

MatthewTighe:
I think you'll also need to rebase on the feature branch since I accepted your other PR first. Sorry!

PR #702, opened 2019-03-14
Validate filesystem names

luongthomas:
**Describe the pull request**

Validate filesystem names during creation to avoid bad characters, including FAT special characters.  
Used this as a reference on what characters to filter out:
https://stackoverflow.com/questions/4814040/allowed-characters-in-filename

Screen Shot 2019-03-14 at 11 20 34 AM

Screen Shot 2019-03-14 at 11 20 21 AM

**Link to relevant issues**
N/A

### DIFF
--- a/app/src/main/java/tech/ula/utils/ValidationUtility.kt
+++ b/app/src/main/java/tech/ula/utils/ValidationUtility.kt
@@ -64,16 +64,12 @@ class ValidationUtility {
     }
 
     private fun validateFilesystemNameCharacters(filesystemName: String): Boolean {
-        val usernameRegex = "([a-zA-Z0-9!@#$%^&()_+=,.?<>]{0,50})"
+        val filesystemNameRegex = "([a-zA-Z0-9!@#$%^&()_+=,.?<>]{0,50})"
 
-        val compiledRegex = Pattern.compile(usernameRegex)
+        val compiledRegex = Pattern.compile(filesystemNameRegex)
         val matcher = compiledRegex.matcher(filesystemName)
         val hasValidCharacters = matcher.matches()
-
-        if (hasValidCharacters) {
-            return true
-        }
-        return false
+        return hasValidCharacters
     }
 
     private fun validateUsernameCharacters(username: String): Boolean {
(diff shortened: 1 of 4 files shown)


app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -48,6 +63,19 @@ class ValidationUtility {
         }
     }
 
+    private fun validateFilesystemNameCharacters(filesystemName: String): Boolean {
+        val usernameRegex = "([a-zA-Z0-9!@#$%^&()_+=,.?<>]{0,50})"
```
MatthewTighe: This regex pattern could use a different name imo.

app/src/test/java/tech/ula/utils/ValidationUtilityTest.kt
```diff
@@ -26,6 +26,38 @@ class ValidationUtilityTest {
...
+    @Test
+    fun `Validate fails appropriately if filesystem name is empty`() {
+        val filesystemName = ""
+        credential = validationUtility.validateFilesystemName(filesystemName)
+        assertFalse(credential.credentialIsValid)
+        assertEquals(credential.errorMessageId, R.string.error_filesystem_name)
+    }
+
+    @Test
+    fun `Validate fails appropriately if filesystem name is just a period`() {
```
MatthewTighe: Would add a test for the `..` case as well? Or just add it to this test.

app/src/main/java/tech/ula/utils/ValidationUtility.kt
```diff
@@ -48,6 +63,19 @@ class ValidationUtility {
...
+
+        val compiledRegex = Pattern.compile(usernameRegex)
+        val matcher = compiledRegex.matcher(filesystemName)
+        val hasValidCharacters = matcher.matches()
+
+        if (hasValidCharacters) {
```
MatthewTighe: I think you could just `return hasValidCharacters`.

luongthomas:
Added one test and renamed a few variables

PR #716, opened 2019-03-22
Arabic language update >> With latest updates :)

foodmaher:
# Arabic language translate ...

**values-ar/strings.xml**


foodmaher:
The Arabic language has been updated with the latest changes. Please check the text (especially the text containing the text formatting and the addition of a modern one) and make sure it is valid before making the link. Also, I want to know how to update my fork in a simple way without having to Terminal Because I always have difficulty updating my Fork. **I want it in an easy and clear way**

app/src/main/res/values-ar/strings.xml
```diff
@@ -15,13 +15,17 @@
...
     <string name="button_continue">استمرار</string>
+    <string name="button_advanced_options_show">اظهار الاعدادات المتقدمة</string>
+    <string name="button_advanced_options_hide">اخفاء الاعدادات المتقدمة</string>
+    <string name="button_select_backup_import">حدد ملف النسخة الاحتياطية للستيراد</string>
 
     <!-- Menus -->
     <string name="app_info">معلومات التطبيق</string>
     <string name="stop_app">ايقاف التطبيق</string>
-    <string name="stop_session">قتل الخدمة</string>
+    <string name="stop_session">قتل الخدمة</string>               <!-- We know this says name="kill_service" but it is really stopping a session. -->
```
luongthomas: Our latest English string actually uses `"Stop Session"` now something like `توقف الجلسة` would be better (I used google translate for that 😅).  

Also you can remove the comment `<!-- We know this says name="kill_service" but it is really stopping a session. -->` since it doesn't use "kill service" anymore

foodmaher:
@luongthomas There are two other issues that I will not be able to resolve because Android Studio is not available. I do edition everything on my phone.

There is a space between words
Screenshot_2019-04-02-07-13-42

Take the opposite side in the last field (because there is an English text with Arabic).
Screenshot_2019-04-02-07-13-57

app/src/main/res/values-ar/strings.xml
foodmaher: modified

luongthomas:
> @luongthomas There are other issue that I will not be able to resolve because Android Studio is not available. I do edition everything on my phone.
> 
> Take the opposite side in the last field (because there is an English text with Arabic).



After playing around with it, I found a stackoverflow post that addresses this issue:
https://stackoverflow.com/a/45447164/8105638

You can change the value in line 39 (`hint_vnc_password`) in `app/src/main/res/values-ar/strings.xml` from 

`<string name="hint_vnc_password">VNC كلمة السر</string>`
to
`<string name="hint_vnc_password">\u200FVNC كلمة السر</string>`

the `\u200F` is the Unicode Character 'RIGHT-TO-LEFT MARK' (U+200F) which will force the string to be displayed Right to Left.

luongthomas:
Screen Shot 2019-04-02 at 11 05 21 AM

This is what it looks like on my device.

luongthomas:
Looks great, thanks @foodmaher

Issue #703, opened 2019-03-15
UserLand has entered an illegal state couldn't extract filesystem


Appeee:
After downloading the assets UserLand quits with the message in the title. Clearing the support files and reinstalling the app does not help. Log attached. "proot info: vpid 1: terminated with signal 11"

# Device Information

Device: Sony Xperia Z3
Android Version: Oreo 8.1
Kernel: 3.4.0-lineage-g784c3f5b 
UserLAnd Version 2.3.8  
The device is rooted with Magisk 18.1
Running CarbonRom CR-6.1 - 20190311-1839
[PRoot_Debug_Log.log](https://github.com/CypherpunkArmory/UserLAnd/files/2970292/PRoot_Debug_Log.log)

# Additional context

MatthewTighe:
We recently discovered our download process can cause files to become corrupted, as we are currently downloading our assets directly from a git tree. We're working on an upgrade to the process that will hopefully ship before the end of next week that will fix issues like these. My first assumption is that this is what is causing your issue.

Any other thoughts @corbinlc ?

lucipher666:
Im getting the same message on my alcatel a574bl , Android version 7.1.1, Kernal 3.10.49-g9d0253d unrooted

rbthomp:
I'd just like to add that I'm getting the same error, but mine originated after restarting the phone several times. After words I couldn't get my session to start. I then proceeded to remove the filesystem and recreate it when I received the error. I have had this issue on 4 Galaxy S6 Android version 7.0. for whatever reason debug logging dose not create the logging file.

Appeee:
Hi there,

some progress might have arose cause now it completes the downloads, copies the assetts and only after verifying it pops up the same error message. Hope it helps. A.

Areahints:
Similar issue. But mine was working now suddenly it never passes <illegal state>

MatthewTighe:
Do you have enough internal storage space to extract a filesystem? You should plan to have at least 1GB free. We realized recently that not the process will still be attempted even without the requisite space and fail without a useful error message. We're working on correcting that.

Areahints:
I'm just curious but do all the Linux distributions require 1GB minimum?

MatthewTighe:
No, but they will quickly reach that size as soon as they are used for anything. Even `apt-get update && apt-get upgrade` tend to balloon size substantially.

pferdefreund:
[...]

Issue #692, opened 2019-03-13
Bug report - Filesystem extraction error in the newest version


octalfish:
# Describe the bug

Getting the classic extraction failure error in the newest version (Not beta) downloaded from play store.

# Steps to reproduce the behavior:

1. Open the app
2. Click on any of the available distros (
3. Wait till the "~80mb" installation completes

4. Then it says that it's "Fetching asset Lists" for a while and crashes with an error message that says "failed to install filesystem".

# Screenshots

![IMG_20190313_131356](https://user-images.githubusercontent.com/45511164/54261497-ebca7800-4591-11e9-9362-f79e07e8ea62.JPG)

![IMG_20190313_131339](https://user-images.githubusercontent.com/45511164/54261517-f6850d00-4591-11e9-9881-7b86e97272a7.JPG)

# Device Information

Device: [vivo Y55L].  
Android Version: [Lollipop 5.1.1].  
UserLAnd Version [2.3.8]
The device is neither rooted nor is running LineageOS

# Additional context

Here's what the proot log says:

[H[JWARNING: linker: /data/data/tech.ula/files/1/support/.proot: unused DT entry: type 0x1d arg 0x5ff
WARNING: linker: libtalloc.so.2: unused DT entry: type 0x1d arg 0xc2e
CANNOT LINK EXECUTABLE DEPENDENCIES: cannot locate symbol "__register_atfork" referenced by "libtalloc.so.2"...

And yes, i have tried re-installing assets and the app itself several times now but the problem still persists

rbthomp:
I'm getting the same message on several Galaxy S6 phones on 2.3.8. So far I've seen it on 3 different phones. 

Everything was running fine till after a couple phone reboots. I now can't even create a new file system after deleting the old one. I don't have any debugging logs to see what is going on, because even after enabling debug logging the log file never gets created. 

I'm happy to help collect information on the issue, but can't get logging to work.

corbinlc:
Is this still happening on the latest release?

TEHONX13:
If I have also had this same problem I have tried everything I have found but none of this works and if this error continues to persist I hope that it will be solved soon my device is a Huawei Y7 pro

corbinlc:
Closing as part of a repository cleanup. If this issue is still relevant, feel free to reopen it or file a fresh one.

Issue #685, opened 2019-03-10
Error setting up filesystem


wpwoodjr:
# Describe the bug
During initial setup of Ubuntu, Firefox, and Git I receive this error and cannot proceed:
![Screenshot 2019-03-10 at 2 26 31 PM](https://user-images.githubusercontent.com/6633152/54089967-44f19c00-4345-11e9-95f1-5f45f87ee976.png)

# Steps to reproduce the behavior:

1. Install UserLAnd on Pixelbook
2. Attempt to setup Ubuntu

# Screenshots

![Screenshot 2019-03-10 at 2 26 31 PM](https://user-images.githubusercontent.com/6633152/54089967-44f19c00-4345-11e9-95f1-5f45f87ee976.png)

# Device Information

Device: Pixelbook i7
Android Version: Pie 9
UserLAnd Version: 2.3.8
Whether the device is rooted.  No
Whether the device is running LineageOS No

# Additional context

None

JasonsDoomGitHub:
Try installing Ubuntu by itself. Then within the ssh install your desired packages. 
I'm guessing its relaying the issues because each option in the distribution is basically like a distro itself. And UserlAnd only let's you use One app at a time. 
Like I mentioned try installing the Ubuntu Distribution by itself and then download your desired packages within the ssh terminal it should work perfectly fine. Or it couple be that the app itself isn't designed yet, to work with OS.v9 
Idk just a suggestion, I'm not saying it will fix the problem but hey either way what is there to lose if u don't try. 
Yours truly 
-jasonsdoom

corbinlc:
Function not implemented is a PRoot error.  We will need a PRoot log.  In UserLAnd, please go to settings, enable PRoot logging and set the verbosity level to 9.  The try the install again.  After it fails, please send us the generated log file.

wpwoodjr:
I tried to get a debug log but it never gets past "Verifying files".  Does debug log require an external SD card?

MatthewTighe:
It's possible that the logging was overwriting the UI output. Check if `/sdcard/PRoot_Debug_Log` exists.

corbinlc:
Closing as part of a repository cleanup. If this issue is still relevant, feel free to reopen it or file a fresh one.

Issue #634, opened 2019-02-13
Failed to extract filesystem (Beta version)


foodmaher:
# Failed to extract filesystems

- I tried to scan and reinstall it but it also does not work
- I also tried to clear support files but it does not work too

# Describe how your task behaves in a regular Linux environment
- The problem is only related to userland

# Additional documents
![screenshot_2019-02-13-19-13-13](https://user-images.githubusercontent.com/27397889/52731019-64efa300-2fc5-11e9-839a-3866e162f743.png)
**Edit:** It seems that the problem appears only in the trial version because I downloaded the default version and work well Did you continue to use the trial version because it has the latest features or use the default version

corbinlc:
Can you create a new filesystem and try again?  We updated PRoot after you ran into this issue.

foodmaher:
@corbinlc Yes, the trial version is running after the update where Debian has been successfully installed..

PR #722, opened 2019-03-25
Download assets from github releases

MatthewTighe:
**Describe the pull request**
This PR switches our download implementation to use the releases section of Github. This is to prevent asset corruption that was found when downloading assets directly from the git tree. 

The new flow queries the Github releases API to find endpoints for the assets list text file and the assets themselves, which are now released as tar.gz archives. Therefore:

1. Query API to find assets list text file
2. Read assets list directly into memory
3. Query latest release version
4. Download newest archives as necessary
5. Unpack newest archives as necessary
6. Verify unpacked assets match assets list

A good to place to start reviewing could be the changes to the session FSM. All steps from retrieving the assets list to verifying assets have been touched.

**Link to relevant issues**
This PR is intended to fix many of the issues we've seen cropping up around download issues, filesystem, and asset corruption.

Potentially affects #703, #692, #685, #634, #566, #486, #462, #461, #409, #380, #364.


app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -85,17 +82,29 @@ class DefaultPreferences(private val prefs: SharedPreferences) {
 }
 
 class AssetPreferences(private val prefs: SharedPreferences) {
-    private fun String.addTimestampPrefix(): String {
-        return "timestamp-" + this
+
+    private val versionString = "version"
```
luongthomas: It appears this string is a constant so should we change the variable name to all caps to signify that it won't change?

I was reading about it in this stackoverflow post: 
https://stackoverflow.com/questions/44038721/constants-in-kotlin-whats-a-recommended-way-to-create-them

It may help readability, particularly when using them in the string templates (for example line 100)

MatthewTighe: It looks like that's not actually the case for private variables. I think that convention is meant for global and static constants.
https://discuss.kotlinlang.org/t/naming-convention-for-private-constants/8871

Renaming it according to your suggestions makes the linter complain.

MatthewTighe: image

PR #736, opened 2019-03-28
Parse download failures into human-readable strings.

MatthewTighe:
**Describe the pull request**

This will allow our users to have a better understanding of why the download manager fails downloads.

**Link to relevant issues**
N/A


luongthomas:
A suggestion to improve it is to rearrange the status codes to be in ascending/descending order but it's also fine as is

MatthewTighe:
I seem where you're coming from on the order, but I just followed the order they're list in on the [DownloadManager docs page](https://developer.android.com/reference/android/app/DownloadManager).

Issue #617, opened 2019-02-12
Requesting Feature : Extra Keyboard Row


Chidhambararajan:
I have been using termux for a long time , but I recently shifted to UserLand because of its ability to give root access, but I find a small draw back in this app . 

The extra keyboard row feature in termux has 2 rows which is really useful but UserLand has only 1 such row.. is there a way to add the same ??

Termux screen shot highlighting the same
![](https://i.imgur.com/HknqtBz.jpg)

UserLand screenshot with the absence of the 2nd row
![](https://i.imgur.com/0ASVYgo.jpg)

Kindly add the same feature or explain the way to add the same

Thanks In Advance

EnderNightLord-ChromeBook:
You Should Try ConnectBot With UserLAnd It Works Very Well.

Herzogs:
sorry the question does not exist a way to assign the keys that we want? for example in Termux there is to edit the file termux.properties

ch2925:
write your termux.properties here:/host-rootfs/data/data/tech.ula/files/home/.termux/termux.properties

xamarama:
"Permission denied". /host-rootfs

Herzogs:
1. `cd /host-rootfs/data/data/tech.ula/files/home/`
2. `mkdir -p .termux`
3. `cd .termux`
4. `busybox vi termux.properties`
5. Into the file put: `extra-keys=[['ESC','TAB','CTRL','ALT','UP','DOWN','LEFT','RIGHT']]` or keys what you use

xamarama:
Thanks. It is work

corbinlc:
We added an extra row and @Herzogs provided how to customize it.  I am closing this now

PR #743, opened 2019-04-01
Change default extra-keys bar in terminal

luongthomas:
**Describe the pull request**

For the UserLAnd terminal, this PR adds a second row in the extra-keys bar (which sits above the keyboard) that includes the following keys: TAB, HOME, END, PGUP, PGDOWN, LEFT and RIGHT.

It is important to note that for any user that has configured their own extra-keys layout, their layout will still be in effect. 👍 

I've changed my extra-keys layout according to this comment here:
https://github.com/CypherpunkArmory/UserLAnd/issues/617#issuecomment-477759175

Previously there was an issue when adding a second row.  If a swipe action occurred (for example to swipe left to go from the first to the second row) the initial touch down action would count as an key press event.  This PR will fix this issue.  When swiping, there will be no key press events submitted.

Also this PR reduces the height of the extra-keys bar by half to take up less space.

**Link to relevant issues**
#617

### DIFF
--- a/termux-app/terminal-term/src/main/java/com/termux/app/ExtraKeysView.java
+++ b/termux-app/terminal-term/src/main/java/com/termux/app/ExtraKeysView.java
@@ -36,6 +36,12 @@ public final class ExtraKeysView extends GridLayout {
     private static final int INTERESTING_COLOR = 0xFF80DEEA;
     private static final int BUTTON_PRESSED_COLOR = 0x7FFFFFFF;
 
+    /**
+     *  Variables used when swiping the extra keys row. Prevent a KEY_PRESS_DOWN when swiping
+     */
+    private float actionDownPosition;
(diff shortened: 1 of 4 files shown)


luongthomas:
[Extra keyboard Row (Include up and down keys)](https://trello.com/c/0GQWXvaZ/591-extra-keyboard-row-include-up-and-down-keys)

luongthomas:
> I'm worried that users unfamiliar with Termux won't be aware that swiping gives them access to more keys/options. I think it might be more useful to have both rows appear simultaneously, like they do in the OP of #617. You can achieve this behavior by changing `TermuxPreferences#151` to to include two arrays instead of just one. This would allow you to remove the logic around created a second array.
> 
> What do you think about that change?

I think that's a great idea!
I've implemented it with my latest commit.  


Also here's a picture of what it looks like now.
Screen Shot 2019-04-02 at 10 30 28 AM

PR #757, opened 2019-04-10
Download generation fix

### LINKED ISSUE #753: Bug report - userland does not start without working network
# Describe the bug

Since last update of UserLand it does not start without having a working network-connection. The last update was 2 days ago from google-playstore. I use userland often when going to work with railroad and there i have no working network. So i can't start Userland in the train and work with it.

# Steps to reproduce the behavior:

1 I updated userland from google playstore.
2 i stopped networking on my samsung A5(6) because i went to train.
3 i started userland and it brings an error that neworking is missing.
4 when i hat neworking again it starts without any problems and works fine.
5 When i let it running after i started it with networking and i stop networking without stopping userland, it works fine too. My problem is, that i can only start it with a working network-connection.

# Screenshots

# Device Information
Samsung A5 (6) with Android 7.
  
UserLAnd Version 2.5.3
Device is not rooted and runs Android 7.
I tried, to clear the cache of the app and i tryed clear all support files but the problem exists further.
Please excuse my bad engish. I come from Germany and learned it about 45 years ago.
Greetings from Germany
Klaus Schaefer

MatthewTighe:
**Describe the pull request**

Previously, we were allowing network failures during download generation to propagate regardless of whether assets were already downloaded. Network attempts are made every time a session is starting to see if there are updates to assets. This PR will allow failed network attempts to be ignored if the assets being checked for updates exist already in support directories. This will allow users to use UserLAnd offline once they have completed the setup process.

**Link to relevant issues**
Closes #753

### DIFF
--- a/app/src/test/java/tech/ula/model/repositories/AssetRepositoryTest.kt
+++ b/app/src/test/java/tech/ula/model/repositories/AssetRepositoryTest.kt
@@ -19,6 +19,7 @@ import tech.ula.utils.ConnectionUtility
 import java.io.File
 import java.io.IOException
 import java.lang.IllegalStateException
+import java.net.UnknownHostException
 
 @RunWith(MockitoJUnitRunner::class)
 class AssetRepositoryTest {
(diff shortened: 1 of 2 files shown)


app/src/test/java/tech/ula/model/repositories/AssetRepositoryTest.kt
```diff
@@ -226,6 +265,58 @@ class AssetRepositoryTest {
...
+        val assetFile = File("$distDirectory/${distAsset.name}")
+        distDirectory.mkdirs()
+        assetFile.createNewFile()
+
+        whenever(mockAssetPreferences.getLatestDownloadVersion(repo))
+                .thenReturn(lowVersion)
+        runBlocking {
+            whenever(mockGithubApiClient.getLatestReleaseVersion(repo))
+                    .thenThrow(UnknownHostException())
+                    .thenThrow(UnknownHostException())
```
luongthomas: This appears to be an extra line

MatthewTighe: The mock throws once when the stub is during the invocation to `lastDownloadedVersionIsUpToDate(repo)` at line 50, and then again when it is called during the invocation to `lastDownloadedFilesystemVersionIsUpToDate(repo)` at line 71.

PR #764, opened 2019-04-12
Display storage capacity warning when low on storage

luongthomas:
**Describe the pull request**
When the available storage on a device is less than 1 gb, and the user attempts to start a session or an app, warn them that Userland works best when there is at least 1gb of available space.  This will still allow them to continue with starting the session/app if they want to.

**Link to relevant issues**


luongthomas:
Closed in favor of #770

PR #770, opened 2019-04-16
Add Storage capacity warning when device is low in memory

luongthomas:
**Describe the pull request**

When the available storage on a device is less than 1 gb but greater than 250MB, and the user attempts to extract a filesystem, warn them that Userland works best when there is at least 1gb of available space. This will still allow them to continue with starting the session/app if they want to.  

But if there is less than 250 MB of storage available, an error message will display telling them to free up some memory and the app will send an illegal state. 

Screen Shot 2019-04-15 at 4 31 02 PM
)

Screen Shot 2019-04-15 at 4 33 04 PM
)

**Link to relevant issues**

### DIFF
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -400,6 +400,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
     private fun handleUserInputState(state: UserInputRequiredState) {
         acraWrapper.putCustomString("Last handled user input state", "$state")
         return when (state) {
+            is LowStorageAcknowledgementRequired -> {
+                displayLowStorageDialog()
+            }
             is FilesystemCredentialsRequired -> {
                 getCredentials()
             }
@@ -478,6 +481,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
             }
             is FailedToClearSupportFiles -> {
                 getString(R.string.illegal_state_failed_to_clear_support_files)
+            }
+            is InsufficientAvailableStorage -> {
+                getString(R.string.illegal_state_insufficient_storage)
             }
         }
         displayIllegalStateDialog(reason)
@@ -592,6 +598,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
             is CopyingDownloads -> {
                 val step = getString(R.string.progress_copying_downloads)
                 updateProgressBar(step, "")
+            }
+            is VerifyingAvailableStorage -> {
+                killProgressBar()
             }
             is FilesystemExtractionStep -> {
                 val step = getString(R.string.progress_setting_up_filesystem)
@@ -716,6 +725,12 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
             viewModel.handleUserInputCancelled()
         }
         customDialog.show()
+    }
+
+    private fun displayLowStorageDialog() {
+        displayGenericErrorDialog(this, R.string.alert_storage_low_title, R.string.alert_storage_low_message) {
+            viewModel.lowAvailableStorageAcknowledged()
+        }
     }
 
     private fun getServiceTypePreference() {
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -19,6 +19,7 @@ import android.net.wifi.WifiManager
 import android.os.Build
 import android.os.Bundle
 import android.os.Environment
+import android.os.StatFs
 import android.support.design.widget.TextInputEditText
 import android.support.v4.content.LocalBroadcastManager
 import android.support.v7.app.AppCompatActivity
@@ -124,6 +125,7 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
 
         val busyboxExecutor = BusyboxExecutor(filesDir, Environment.getExternalStorageDirectory(), DefaultPreferences(defaultSharedPreferences))
         val filesystemUtility = FilesystemUtility(filesDir.path, busyboxExecutor)
+        val storageUtility = StorageUtility(StatFs(Environment.getExternalStorageDirectory().path))
 
         val downloadManager = getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager
         val downloadManagerWrapper = DownloadManagerWrapper(downloadManager)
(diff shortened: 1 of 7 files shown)


luongthomas:
[Notify user if there isn't enough space on device to install a filesystem](https://trello.com/c/tN7QyoVE/660-notify-user-if-there-isnt-enough-space-on-device-to-install-a-filesystem)

AndrewScibek:
Should we be crashing the app if they dont have enough memory? That doesn't make much sense

corbinlc:
We shouldn't crash, as there is no error in the flow and a crash log won't help us as developers.  We should give them that message and have it just go back to the page of the app they were on.  It shouldn't let them progress without freeing up space, but it shouldn't crash.

MatthewTighe:
Crashing mirrors failure mode behaviors in other places in the app. I told Thomas that could be the functionality implemented now, and that we would go back through it as part of https://trello.com/c/b1veJqAS

app/src/main/java/tech/ula/MainActivity.kt
```diff
@@ -593,6 +599,9 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection,
                 val step = getString(R.string.progress_copying_downloads)
                 updateProgressBar(step, "")
             }
+            is VerifyingAvailableStorage -> {
+                killProgressBar()
```
MatthewTighe: I think that the progress bar should still be visible during this state. That way, in the successful case, the progress bar won't disappear between between the asset verification and filesystem extraction steps. While you're in there, can you reorder the `when` statement to reflect the actual order of steps? I.E verifying filesystem should be before extraction.

app/src/main/java/tech/ula/model/state/SessionStartupFsm.kt
```diff
@@ -339,5 +364,7 @@ data class AssetDownloadComplete(val downloadAssetId: Long) : SessionStartupEven
 object SyncDownloadState : SessionStartupEvent()
 object CopyDownloadsToLocalStorage : SessionStartupEvent()
 data class VerifyFilesystemAssets(val filesystem: Filesystem) : SessionStartupEvent()
+object VerifyAvailableStorage : SessionStartupEvent()
+object VerifyAvailableStorageComplete : SessionStartupEvent()
```
MatthewTighe: I think `VerifyAvailableStorageComplete` and the associated logic could be removed from the FSM. Rather than wait for that event here and emit the success state for this stage, I think the VM could directly submit the `ExtractFilesystem` event when the `VM#lowAvailbleStorageAcknowledged` callback is made.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -46,12 +47,13 @@ fun arePermissionsGranted(context: Context): Boolean {
                     Manifest.permission.WRITE_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED)
 }
 
-fun displayGenericErrorDialog(activity: Activity, titleId: Int, messageId: Int) {
+fun displayGenericErrorDialog(activity: Activity, titleId: Int, messageId: Int, callback: (() -> Unit)? = null) {
```
MatthewTighe: I think the `callback` parameter could better follow idiomatic Kotlin by using an empty function default, `{}` and a non-nullable type.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -66,6 +68,13 @@ fun getBranchToDownloadAssetsFrom(assetType: String): String {
     }
 }
 
+fun getAvailableStorageInMB(): Long {
```
MatthewTighe: AFAIK, there's no way to mock a top-level function like this, which I would guess is what you were talking about having trouble testing. Try inserting it into a class instead. It would be _kind of_ stretching the defintion of the filesystem utility, but you could add it there if you wanted, especially if you rephrased it a bit to something like `FilesystemUtility#spaceRequiredForNewFilesystemIsAvailable`. I can describe this more offline but I'm tired of typing.

app/src/test/java/tech/ula/viewmodel/MainActivityViewModelTest.kt
```diff
@@ -658,11 +658,22 @@ class MainActivityViewModelTest {
     }
 
     @Test
-    fun `Submits ExtractFilesystem when observing FilesystemAssetVerificationSucceeded`() {
+    fun `Submits VerifyAvailableStorage when observing FilesystemAssetVerificationSucceeded`() {
```
MatthewTighe: I'd like to see more tests for the rest of the states you introduced.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -68,11 +68,14 @@ fun getBranchToDownloadAssetsFrom(assetType: String): String {
     }
 }
 
-fun getAvailableStorageInMB(): Long {
-    val bytesInMB = 1048576
-    val stat = StatFs(Environment.getExternalStorageDirectory().path)
-    val bytesAvailable = stat.blockSizeLong * stat.availableBlocksLong
-    return bytesAvailable / bytesInMB
+class StorageUtility(statFs: StatFs) {
```
MatthewTighe: You can declare `statFs` as a member property of the class in the constructor by just doing `class StorageUtility(private val statFs: StatFs)`

app/src/main/res/values/strings.xml
```diff
@@ -120,6 +120,7 @@
     <string name="progress_downloading">Downloading required assets&#8230;</string>
     <string name="progress_downloading_out_of">%1$d out of %2$d complete</string>
     <string name="progress_copying_downloads">Copying downloaded files to local storage&#8230;</string>
+    <string name="progress_verifying_sufficient_storage">Verifying sufficient storage available</string>
```
MatthewTighe: Would you add the trailing ellipsis character to this string? `&#8230;`

app/src/test/java/tech/ula/model/state/SessionStartupFsmTest.kt
```diff
@@ -720,6 +722,42 @@ class SessionStartupFsmTest {
...
+
+        whenever(mockStorageUtility.getAvailableStorageInMB()).thenReturn(150)
+        runBlocking { sessionFsm.submitEvent(VerifyAvailableStorage, this) }
+
+        verify(mockStateObserver).onChanged(VerifyingSufficientStorage)
+        verify(mockStateObserver).onChanged(VerifyingSufficientStorageFailed)
+    }
+
+    @Test
+    fun `State is not  if device has greater than 1GB of storage after VerifyingAvailableStorage`() {
```
MatthewTighe: This test name looks like it's missing a word.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -46,12 +47,13 @@ fun arePermissionsGranted(context: Context): Boolean {
...
 }
 
-fun displayGenericErrorDialog(activity: Activity, titleId: Int, messageId: Int) {
+fun displayGenericErrorDialog(activity: Activity, titleId: Int, messageId: Int, callback: (() -> Unit) = {}) {
     AlertDialog.Builder(activity)
             .setTitle(titleId)
             .setMessage(messageId)
             .setPositiveButton(R.string.button_ok) {
                 dialog, _ ->
+                callback?.let { it() }
```
MatthewTighe: You can remove the nullable guard on the callback now.

PR #813, opened 2019-05-07
Automate github releases.

MatthewTighe:
**Describe the pull request**
This PR will add a step to our automated deployment that will also release the apk to Github as a new release with the same changelogs as what is being sent to the playstore.

**Link to relevant issues**
N/A


.circleci/config.yml
```diff
@@ -79,6 +79,11 @@ reference:
       name: Decode Android key store
       command: echo $ANDROID_KEYSTORE | base64 -di > keystore.jks
 
+  export_github_release_token: &export_github_release_token
+    run:
+      name: Export Github release key
+      command echo 'export GITHUB_RELEASE_TOKEN="$GITHUB_RELEASE_TOKEN"' >> $BASH_ENV
```
luongthomas: I think this is missing a colon at the end of `command` which is leading to CircleCI not building

MatthewTighe: Yep, you were right

PR #819, opened 2019-05-09
Fix broken constraint layouts.

MatthewTighe:


MatthewTighe:
Good find, fixed that.

Issue #762, opened 2019-04-11
"Command failed with: 126"


lapsio:
So I tried to install UserLAnd in Work Profile and got following crash when filesystem was setting up:

![Screenshot_20190411-164218](https://user-images.githubusercontent.com/1054079/55967270-4f6fc000-5c7a-11e9-8a82-b6802fd9eb1c.png)

And then every attempt to run session ends up with:

![Screenshot_20190411-164659](https://user-images.githubusercontent.com/1054079/55967313-62829000-5c7a-11e9-9b99-48f6f0a35bad.png)

Where is mentioned error log placed?

# Describe the bug

Initial session setup crashes when app is installed in Work Profile.

# Steps to reproduce the behavior:

1. Set up Android Work Profile container.
2. Install UserLAnd in Work Profile container
3. Create new filesystem and session
4. Try to launch session

# Screenshots

![Screenshot_20190411-164218](https://user-images.githubusercontent.com/1054079/55967270-4f6fc000-5c7a-11e9-8a82-b6802fd9eb1c.png)

![Screenshot_20190411-164659](https://user-images.githubusercontent.com/1054079/55967313-62829000-5c7a-11e9-9b99-48f6f0a35bad.png)

# Device Information

Device: Asus ZS570KL.  
Android Version: 8.0.  
UserLAnd Version .  2.5.4
Whether the device is rooted.  no
Whether the device is running LineageOS. no

lapsio:
added headers

lapsio:
Itried to clean support files. After repeating procedure I didn't get "Permission denied" error. Only that "126" one.

lapsio:
Okay I found log file, it contains 1 line:

    sh: support/execInProot.sh: Permission denied

MatthewTighe:
Do things work normally while if the app is installed in a regular profile?

lapsio:
> Do things work normally while if the app is installed in a regular profile?

I do have UserLAnd also installed in regular profile. Calling it "working normally" would be a bit of stretch but yes, it's _usually_ functional. At least session starts properly.

lapsio:
I don't know how do you evoke scripts but maybe you have some hardcoded paths? Work Profile has different app data paths:

On top terminal in work profile, at the bottom terminal in normal profile. All paths typically containing "0" like `/storage/emulated/0` etc contain different numbers eg. 10

[...]

Issue #607, opened 2019-02-09
Can Not use on 2. Account


Proxtx:
I am using Android 9 and it returns an error

EnderNightLord-ChromeBook:
Please Use The Issue Template Otherwise This Post Will Be Closed.

MatthewTighe:
Please update this issue to follow the issue template provided [here](https://github.com/CypherpunkArmory/UserLAnd/blob/master/.github/ISSUE_TEMPLATE/bug_report.md).

PR #837, opened 2019-05-17
Add support for multiple Android accounts.

corbinlc:
**Describe the pull request**

Makes it so extraction, running the server and connecting the client all work on a 2nd account.
Still some oddity when running say `sudo apt-get update` but I think that is going to take some PRoot debugging to fix.

**Link to relevant issues**

#607 #762 probably others

### DIFF
--- a/app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
+++ b/app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
@@ -140,7 +140,7 @@ class BusyboxExecutor(
 class BusyboxWrapper {
     // For basic commands, CWD should be `applicationFilesDir`
     fun addBusybox(command: String): List<String> {
-        return listOf("support/busybox", "sh", "-c", command)
+        return listOf("support/busybox", "sh") + command.split(" ")
     }
 
     fun getBusyboxEnv(filesDir: File): HashMap<String, String> {
@@ -154,8 +154,7 @@ class BusyboxWrapper {
 
     // Proot scripts expect CWD to be `applicationFilesDir/<filesystem`
     fun addBusyboxAndProot(command: String): List<String> {
-        val commandWithProot = "support/execInProot.sh $command"
-        return listOf("support/busybox", "sh", "-c", commandWithProot)
+        return listOf("support/busybox", "sh", "support/execInProot.sh")  + command.split(" ")
     }
 
     fun getProotEnv(filesDir: File, filesystemDir: File, prootDebugLevel: String, externalStorageDir: File): HashMap<String, String> {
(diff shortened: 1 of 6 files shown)


termux-app/terminal-term/src/main/java/com/termux/app/TermuxActivity.java
```diff
@@ -150,8 +152,8 @@ public void onReceive(Context context, Intent intent) {
 
     void checkForFontAndColors() {
         try {
-            @SuppressLint("SdCardPath") File fontFile = new File("/data/data/tech.ula/files/home/.termux/font.ttf");
-            @SuppressLint("SdCardPath") File colorsFile = new File("/data/data/tech.ula/files/home/.termux/colors.properties");
+            @SuppressLint("SdCardPath") File fontFile = new File(this.getFilesDir().getAbsolutePath() + "/home/.termux/font.ttf");
```
MatthewTighe: These lint suppressions might not be necessary now.

termux-app/terminal-term/src/main/java/com/termux/app/TermuxService.java
```diff
@@ -50,10 +50,10 @@
 
     /** Note that this is a symlink on the Android M preview. */
     @SuppressLint("SdCardPath")
-    public static String FILES_PATH = "/data/data/tech.ula/files";
-    public static final String SUPPORT_PATH = FILES_PATH + "/support/";
-    public static final String PREFIX_PATH = FILES_PATH + "/usr";
-    public static final String HOME_PATH = FILES_PATH + "/home";
+    public String FILES_PATH;
```
MatthewTighe: Non-static fields should use camelcase naming.

app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
```diff
@@ -154,8 +154,7 @@ class BusyboxWrapper {
 
     // Proot scripts expect CWD to be `applicationFilesDir/<filesystem`
     fun addBusyboxAndProot(command: String): List<String> {
-        val commandWithProot = "support/execInProot.sh $command"
-        return listOf("support/busybox", "sh", "-c", commandWithProot)
+        return listOf("support/busybox", "sh", "support/execInProot.sh") + command.split(" ")
```
MatthewTighe: Is splitting the command necessary now?

MatthewTighe (changes requested):

app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
corbinlc: Yes, the `-c` before treated each arg differently.

corbinlc:
I found and fixed a very old PRoot bug that was manifesting itself when running a 2nd account (no idea why it didn't show up elsewhere).  So, this now should fully work as well as being on the primary account.

MatthewTighe (approved):

PR #841, opened 2019-05-20
Add logic to graphical app startup

luongthomas:
**Describe the pull request**

There was a visual bug that occurred when starting a GUI-only app on an Android 9.0 device.  It should start VNC automatically right after clicking the graphical-only app, instead it shows a dialog with SSH selected.  

This PR will fix the behavior to have VNC automatically selected.

**Link to relevant issues**


luongthomas:
[If graphical app is selected on android 9.0 device, select VNC automatically](https://trello.com/c/7QRgGYVS/675-if-graphical-app-is-selected-on-android-90-device-select-vnc-automatically)

PR #860, opened 2019-05-31
Bump version.

MatthewTighe:


MatthewTighe:
Man sometimes this CI chain is just _too useful_.

PR #868, opened 2019-06-04
Hotfix for deleting filesystem crash.

MatthewTighe:
**Describe the pull request**
A bug was introduced in which we were wrapping basic busybox commands incorrectly. Commands still need the `-c` option sent to `busybox sh` whereas scripts can no longer use it due to the changes made to allow multi-user environments. This PR codifies this discrepancy by introducing a separate function `BusyboxExecutor#executeScript` and calls the appropriate functions in the right places.

**Link to relevant issues**
N/A

### DIFF
--- a/app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
+++ b/app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
@@ -23,10 +23,10 @@ class BusyboxExecutor(
     private val discardOutput: (String) -> Any = { }
 
     fun executeScript(
-        command: String,
+        scriptCall: String,
         listener: (String) -> Any = discardOutput
     ): ExecutionResult {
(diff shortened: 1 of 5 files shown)


app/src/main/java/tech/ula/utils/BusyboxExecutor.kt
```diff
@@ -102,7 +116,10 @@ class BusyboxExecutor(
 
     suspend fun recursivelyDelete(absolutePath: String): ExecutionResult = withContext(Dispatchers.IO) {
         val command = "rm -rf $absolutePath"
-        return@withContext executeCommand(command)
+        val listener: (String) -> Unit = { line ->
+            Log.e("delete", line)
+        }
```
corbinlc: remove this debug code

MatthewTighe: done

PR #872, opened 2019-06-05
Hotfix notification crash

luongthomas:
**Describe the pull request**

This PR fixes the issues for devices that crash upon opening a userland terminal.  This behavior was noticed on devices with languages other than English.  

The problem came from a mismatch of notification channel IDs, the Userland "app" notification was changing its notification channel ID when the language was set to something else besides english.  This causes a crash because the terminal couldn't find the channel it was looking for.  

The solution is just to set the channel ID as a hardcoded string that matches on the "app" side and the terminal side.

**Link to relevant issues**

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,8 +17,8 @@ android {
         applicationId "tech.ula"
         minSdkVersion 21
         targetSdkVersion 28
-        versionCode 66
-        versionName "2.5.12"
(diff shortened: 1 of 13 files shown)


luongthomas:
[Fix notifications crash](https://trello.com/c/QZHHWqxa/713-fix-notifications-crash)

PR #909, opened 2019-06-19
Fix compilation issues while targeting Q.

MatthewTighe:
This PR fixes compilation issues that occurred while targeting Q. They fell into two main categories:
- Navigation
- Increased number of correctly annotated nullable JVM types

I also took the time to do some general cleanup while I was rooting around.

App would not compile targeting SDK 29.

Probably with the navigation stuff in `nav_graph.xml` and the various fragments.

Yes, by navigating around and starting a session.

Allows us to target Q and remain up-to-date.


app/src/main/java/tech/ula/utils/QWarningHandler.kt
```diff
@@ -28,7 +28,8 @@ class QWarningHandler(private val prefs: SharedPreferences, private val ulaFiles
     }
 
     private fun userHasFilesystems(): Boolean {
-        val hasFilesystems = ulaFiles.filesDir.listFiles().mapNotNull { it.name.toIntOrNull() }.isNotEmpty()
+        val files = ulaFiles.filesDir.listFiles() ?: return false
```
luongthomas: That's neat that you can return in a ternary operation

app/src/test/java/tech/ula/utils/ServerUtilityTest.kt
```diff
@@ -23,15 +23,15 @@ class ServerUtilityTest {
```
luongthomas: This should be removed since the `LogUtility` class was removed

luongthomas: All references to LogUtility in this test file should be removed, that may be the cause of CI failing.

MatthewTighe: Yep, that should fix it.

app/src/main/java/tech/ula/utils/Logger.kt
```diff
@@ -85,12 +92,14 @@ class SentryLogger : Logger {
...
         Sentry.capture(event)
+        Log.e("ILLEGAL_STATE", message)
     }
 
     override fun sendEvent(message: String) {
         val event = EventBuilder()
                 .withMessage(message)
                 .withLevel(Event.Level.ERROR)
         Sentry.capture(event)
+        Log.e("EVENT", message)
```
luongthomas: Looks like this is causing a failing test, complaining that `Log.e` hasn't been mocked in `FilesystemListViewModelTest`

PR #912, opened 2019-06-20
[Fix] VNC full screen not filling up entire screen

luongthomas:
This PR introduces a fix for when the user starts a VNC session and the display doesn't utilize the entire device's screen.

This fix will address that by taking into account the navigation bar of the device.

This will only affect newly created filesystems.  Existing filesystems will have to edit their `~/.vncrc` and change their session resolution.

`UserLAnd/app/src/main/java/tech/ula/utils/AndroidUtility.kt` line 391 where `DeviceDimensions` is defined.

Tested with all available devices in portrait and landscape mode

Now the black bar on the left side of the screen shouldn't be there for newly created filesystems using vnc.

## Before
Screenshot_20190620-081920

## After
Screenshot_20190620-084649

### DIFF
--- a/app/src/main/java/tech/ula/utils/AndroidUtility.kt
+++ b/app/src/main/java/tech/ula/utils/AndroidUtility.kt
@@ -417,7 +417,7 @@ class DeviceDimensions {
         display.getSize(appUsableSize)
         display.getRealSize(realScreenSize)
 
-        return Point(realScreenSize.x - appUsableSize.x, realScreenSize.y - appUsableSize. y)
+        return Point(realScreenSize.x - appUsableSize.x, realScreenSize.y - appUsableSize.y)
     }
 
     private fun checkOrientation(context: Context): String {
--- a/app/src/main/java/tech/ula/utils/AndroidUtility.kt
+++ b/app/src/main/java/tech/ula/utils/AndroidUtility.kt
@@ -397,7 +397,7 @@ class DeviceDimensions {
         const val landscape = "SCREEN_ORIENTATION_LANDSCAPE"
     }
 
-    fun getDeviceDimensions(windowManager: WindowManager, displayMetrics: DisplayMetrics, orientation: Int) {
+    fun saveDeviceDimensions(windowManager: WindowManager, displayMetrics: DisplayMetrics, orientation: Int) {
         val navBarSize = getNavigationBarSize(windowManager)
         windowManager.defaultDisplay.getRealMetrics(displayMetrics)
         height = displayMetrics.heightPixels
@@ -427,7 +427,7 @@ class DeviceDimensions {
         }
     }
 
-    fun getGeometry(): String {
+    fun getScreenResolution(): String {
         return when (height > width) {
             true -> "${height}x$width"
             false -> "${width}x$height"
--- a/app/src/main/java/tech/ula/utils/AndroidUtility.kt
+++ b/app/src/main/java/tech/ula/utils/AndroidUtility.kt
@@ -392,11 +392,6 @@ class DeviceDimensions {
     private var height = 720
     private var width = 1480
 
-    companion object {
-        const val portrait = "SCREEN_ORIENTATION_PORTRAIT"
-        const val landscape = "SCREEN_ORIENTATION_LANDSCAPE"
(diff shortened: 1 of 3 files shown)


luongthomas:
[Make full screen resolution be actually full screen.](https://trello.com/c/S54x5gpq/705-make-full-screen-resolution-be-actually-full-screen)

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -387,13 +389,42 @@ class LocalFileLocator(private val applicationFilesDir: String, private val reso
...
+        val appUsableSize = Point()
+        val realScreenSize = Point()
+        val display = windowManager.defaultDisplay
+        display.getSize(appUsableSize)
+        display.getRealSize(realScreenSize)
+
+        return Point(realScreenSize.x - appUsableSize.x, realScreenSize.y - appUsableSize.y)
+    }
+
+    private fun checkOrientation(orientation: Int): String {
```
MatthewTighe: I think this function could be removed, since the check above looks so similar.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -387,13 +389,42 @@ class LocalFileLocator(private val applicationFilesDir: String, private val reso
...
 
-    fun getDeviceDimensions(windowManager: WindowManager, displayMetrics: DisplayMetrics) {
-        windowManager.defaultDisplay.getMetrics(displayMetrics)
-        width = displayMetrics.widthPixels
+    companion object {
+        const val portrait = "SCREEN_ORIENTATION_PORTRAIT"
+        const val landscape = "SCREEN_ORIENTATION_LANDSCAPE"
+    }
+
+    fun getDeviceDimensions(windowManager: WindowManager, displayMetrics: DisplayMetrics, orientation: Int) {
```
MatthewTighe: I think this function could use a slightly different name. `get` to me implies it will return the dimensions.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
```diff
@@ -387,13 +389,42 @@ class LocalFileLocator(private val applicationFilesDir: String, private val reso
...
 class DeviceDimensions {
-    private var width = 720
-    private var height = 1480
+    private var height = 720
+    private var width = 1480
 
-    fun getDeviceDimensions(windowManager: WindowManager, displayMetrics: DisplayMetrics) {
-        windowManager.defaultDisplay.getMetrics(displayMetrics)
-        width = displayMetrics.widthPixels
+    companion object {
```
MatthewTighe: It looks to me like this can just be private member values, since they aren't being accessed statically anywhere.

app/src/test/java/tech/ula/utils/DeviceDimensionsTest.kt
```diff
@@ -1,5 +1,6 @@
 package tech.ula.utils
```
MatthewTighe: I'd like to see a test or two that ensured correct behavior from `getNavigationBarSize`

app/src/main/java/tech/ula/utils/AndroidUtility.kt
MatthewTighe: Alternatively, it would be more functional if you left the `get` language and had the function return the geometry as needed, so that there wasn't intermediate state left around.

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I'll just move these into the function since they're only used once

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: I changed `get` to `save`, I think that may be more clear

app/src/main/java/tech/ula/utils/AndroidUtility.kt
luongthomas: Removed this function

PR #931, opened 2019-07-08
 Fix bug that caused landing page to appear on rotation

MatthewTighe:
Fixes a recently introduced bug that would cause navigation to the landing page anytime `MainActivity` was recreated.

Statically setting the nav graph up causes this bug, because `startDestination` is set statically and then altered dynamically, instead of just being set the first time dynamically.

In any of the 3 files.

Yes, by altering the preference and testing lifecycle/navigation.

Eliminates an annoying bug.

### DIFF
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -31,7 +31,6 @@ import android.widget.Toast
 import androidx.navigation.NavController
 import androidx.navigation.Navigation
 import androidx.navigation.findNavController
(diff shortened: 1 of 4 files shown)


corbinlc:
Is the change to the `.kt` file just refactoring?  It looks like a few things are being moved around, but I don't see anything new there.  So, is the part that mattered to fix this issue all in the `.xml` files?

MatthewTighe:
Yeah, the Kotlin changes were just refactoring. Behavior changes are the xml files.

PR #932, opened 2019-07-08
Fix progress dialog displaying after session has started.

MatthewTighe:
This PR adds a state to the main VM called `WaitingForInput` that is entered when startup state is reset.

There was a bug that's been around for a while that would cause the progress dialog to reappear during rotation after a session has been started. It occurred because of the following flow

- Rotation causes activity destruction and recreation
- Activity subscribes to events from the VM
- Last emitted state from VM is that extraction happened successfully
- Activity thinks this means it's time to start the session
- Tries restarting session and re-displays progress dialog

This is now being avoided by making the VM aware of when sessions are actually activated, so that it's most recent emitted state becomes `WaitingForInput` instead.

`MainActivity` or the VM.

Yes, by starting sessions on existing and new filesystems and rotating afterwards.

Another bug fix.

### DIFF
--- a/app/src/main/java/tech/ula/MainActivity.kt
+++ b/app/src/main/java/tech/ula/MainActivity.kt
@@ -360,6 +360,7 @@ class MainActivity : AppCompatActivity(), SessionListFragment.SessionSelection, 
     }
 
     private fun handleSessionHasBeenActivated() {
+        viewModel.handleSessionHasBeenActivated()
         killProgressBar()
     }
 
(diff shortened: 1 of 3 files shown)


corbinlc:
Thomas added a trello card related to the issue of the progress being redisplayed at rotation a couple of week ago.  Not sure if that is what you are reacting to.  If not, please find it.

MatthewTighe:
Yep, I'm just working through the backlog. The card is in review.

PR #934, opened 2019-07-11
Refactor utils package.

MatthewTighe:
This PR attempts to add clarify to the `utils` package by:
- Giving classes more descriptive name
- Breaking apart the monolithic `AndroidUtility` class into more logical pieces.
- Starting to move classes into subpackages.

`AndroidUtility` had become a dumping ground for any class or method that was either too small to "justify" its own file or just didn't seem to have a place elsewhere. This seemed like a pretty bad practice in retrospect.

Anywhere. I _believe_ this is a 'true' refactor in the sense that all I did was move code to different places. I don't remember changing actual behavior anywhere.

Clean install through `apt-get update`

Speedier future development


app/src/main/java/tech/ula/utils/Extensions.kt
```diff
@@ -32,6 +34,18 @@ fun <A, B> zipLiveData(a: LiveData<A>, b: LiveData<B>): LiveData<Pair<A, B>> {
     }
 }
 
+fun Context.displayGenericErrorDialog(titleId: Int, messageId: Int, callback: (() -> Unit) = {}) {
```
luongthomas: Neat way of writing extensions

luongthomas (approved):
Code looks good, tested on my device too 👍

Yang062 (approved):

PR #939, opened 2019-07-12
Fix iilegal state transition when downloads fail by cancelling downloads.

MatthewTighe:
Fixes a Sentry issue by cancelling downloads when one fails. 

This issue was caused when `AssetDownloadComplete` events were submitted to the session FSM while it was in the `AssetDownloadFailure` state.

`AssetDownloader`

No, because it's hard to manipulate the app into this state. Instead I'll just keep an eye on Sentry.

Less illegal state messages.

### DIFF
--- a/app/src/main/java/tech/ula/utils/AssetDownloader.kt
+++ b/app/src/main/java/tech/ula/utils/AssetDownloader.kt
@@ -75,6 +75,7 @@ class AssetDownloader(
 
         if (downloadManagerWrapper.downloadHasFailed(downloadId)) {
             val reason = downloadManagerWrapper.getDownloadFailureReason(downloadId)
+            downloadManagerWrapper.cancelAllDownloads(enqueuedDownloadIds)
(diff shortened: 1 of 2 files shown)


app/src/main/java/tech/ula/utils/AssetDownloader.kt
```diff
@@ -228,6 +229,10 @@ class DownloadManagerWrapper(private val downloadManager: DownloadManager) {
         }
         return DownloadFailureLocalizationData(R.string.download_failure_reason_not_found)
     }
+
+    fun cancelAllDownloads(downloadIds: Set<Long>) {
+        downloadManager.remove(*downloadIds.toLongArray())
```
luongthomas: What does the asterisk signify here?

MatthewTighe: It's a spread operator. `DownloadManager#remove` takes varargs.

PR #947, opened 2019-07-19
Add happy path end-to-end test.

MatthewTighe:
This PR adds an end-to-end test for happy path usage of UserLAnd.

This should help avoid regressions and allow us to extend into more complicated use cases.

`MainActivityTest`

That's kind of the opposite of the point. I watched it do it's thing though.

Higher app confidence.


MatthewTighe:
`testInstrumentationRunnerArguments clearPackageData: 'true'` in `build.gradle` accomplishes this

PR #957, opened 2019-07-24
Bump version.

MatthewTighe:


app/src/main/java/tech/ula/utils/QWarningHandler.kt
```diff
@@ -24,7 +24,7 @@ class QWarningHandler(private val prefs: SharedPreferences, private val ulaFiles
     }
 
     private fun versionDoesNotMatchRequirement(): Boolean {
-        return prefs.getString(versionKey, "") ?: "" < "v2.6.0"
+        return prefs.getString(versionKey, "") ?: "" < "v2.7.0"
```
luongthomas: I had to try this for myself to see how the less than operation worked with strings, but that's neat.

https://pl.kotl.in/-rDwBaQtX

PR #967, opened 2019-07-26
Need to have a license specifically related to our icon

corbinlc:
Might make this for all graphics, but will have to analyze if that is possible later.


corbinlc:
We just need to push this through, so Richard can say what license our icon is under, so he can use it.

corbinlc:
We have to have it somewhere.  The website is a better place.  I will add a card to move it there.

PR #968, opened 2019-07-26
Fix bug that could cause crashes if app service type was changed while active.

MatthewTighe:
This change moves the persistence of apps service types from the cache to the session that will actually run the app, and is therefore stored in the database. It introduces a `ViewModel` for the `AppDetailsFragment` to make the separation of these responsibilities clearer.

Can't even fix a simple bug without a day's worth of refactoring. For shame.

`AppDetails` fragment or VM.

Yes, by trying it. Also units tests now.

A small, hard to hit bug is now erased.


luongthomas (approved):

Donckl (commented):
#1527 app/src/main/java/tech/ula/model/state/AppsStartupFsm.kt

PR #976, opened 2019-08-01
handle and launch on multiple screens

jlainema:
**Note**

Fixes an issue with view changes starting multiple new sessions.

Launches and migrates sessions to external screens, when they are plugged in (and back to device screen when they are plugged out).

I am venturing into the maddeningly recursive depths of actually doing development on Android (eventually for Android). Having an external screen that works seamlessly is nearly a requirement for this.

Simple change, just look through it all.

Plugging & unplugging a USB monitor to an Android-Q Oneplus 7 developer preview 3. Connecting & disconnecting a Bluetooth keyboard on the same device.

Fullscreen desktop-like experience when external monitors are plugged in while running UserLAnd.


ghost (commented):
multiple screens?

ghost (changes requested):
multiple screens?

ghost (changes requested):
multiple screens

qqqwwweee123-arch:
handle and launch on multiple screen?

MatthewTighe:
Any suggestions on how this could be tested without access to more physical screens?

jlainema:
It should work with wireless displays as well. I think you should be able to cast to pretty much any laptop over wifi if you not have other miracast devices and no thunderbolt capable android device.

In android Q, there would also be a developer option to create a virtual display to a window directly on the device.

maysusan:
Thank

maysusan (approved):

jlainema:
Changed the "desktop mode" commit description, and added FLAG_KEEP_SCREEN_ON for the external display when both hardware keyboard and external display are available.

I would have expected this to keep the external display awake, while allowing the phone screen to sleep when working on the activity on the external display; but at least my device keeps both the phone display and the external display awake (this is probably an issue in Android, need to investigate further eventually).

corbinlc:
I am sorry we never got back to this.  We had a large funding change about the time we were discussion this with you.  I will look at this, this week or next.  It looks like pretty good work.

gyggyg234:
How

gyggyg234 (approved):

jlaufersweiler:
I'd be willing to test this with Samsung DeX + the phone screen mirrored to a second monitor via DisplayLink adapter. Would that be a matter of cloning the branch repo to local storage and then pointing F-Droid at it, or is there a more involved build process?

Aidenmichael (approved):

PR #1001, opened 2019-08-12
Fix launching into VNC sessions

corbinlc:
Fix launching VNC sessions/apps

We broke this in our last release and it is effecting many people.
Matthew will probably rework this and add tests, because he is a nice guy.

AppStartupFsm.kt

Yes.  Checking launching a broken session and a new one and toggling the session type between launches.

They can use graphical sessions again.

### DIFF
--- a/app/src/main/java/tech/ula/model/state/AppsStartupFsm.kt
+++ b/app/src/main/java/tech/ula/model/state/AppsStartupFsm.kt
@@ -114,6 +114,7 @@ class AppsStartupFsm(
 
     private suspend fun setServiceType(appSession: Session, serviceType: ServiceType) = withContext(Dispatchers.IO) {
         appSession.serviceType = serviceType
+        appSession.port = if (serviceType == ServiceType.Ssh) 2022 else 51
         sessionDao.updateSession(appSession)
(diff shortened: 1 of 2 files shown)


app/src/main/java/tech/ula/model/state/AppsStartupFsm.kt
```diff
@@ -156,6 +157,8 @@ class AppsStartupFsm(
         state.postValue(SyncingDatabaseEntries)
         appSession.filesystemId = appsFilesystem.id
         appSession.filesystemName = appsFilesystem.name
+        appSession.serviceType = appSession.serviceType
```
MatthewTighe: This line looks unnecessary.

MatthewTighe:
You should change the PR title. Changelogs are generated based on commits.

corbinlc:
Done

app/src/main/java/tech/ula/model/state/AppsStartupFsm.kt
corbinlc: Yes.  I was restoring the code that was there, more or less, before 2.6.1, but I obviously created something redundant here... lol

PR #1006, opened 2019-08-14
Create end-to-end VNC session regression test.

MatthewTighe:
- End-to-end VNC session regression test.
- Handles server connection numbers in an appropriate place.
- Kills potentially leaked coroutines on application death.

Several regressions have been caused around sessions not keep track appropriately of these numbers, which should never actually change.

`LocalServerManager.kt`

Yes, by creating vnc sessions.

Higher confidence in the app.


corbinlc (approved):

eslam772 (commented):
app/src/main/java/tech/ula/ServerService.kt

PR #1011, opened 2019-08-16
Fix new seccomp filtering issues

corbinlc:
Change what version of the support assets we are fetching to get a new version of proot which solves two newly occurring seccomp filtering issues.

`waitpid` and `chdir` were being blocked on certain version of android

Well really all the changes are in proot

We will actually need some help from users with effected devices to be sure things are fixed, but I have tested that this does not seem to break anything.

Users with effected devices should actually be able to user UserLAnd now.

Block chdir, WTF was someone thinking

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ reference:
   android_config: &android_config
     working_directory: *workspace
     docker:
-      - image: circleci/android:api-28-ndk
+      - image: circleci/android@sha256:fa7a00c75f4b28cc4f2a15a382fb76e830d62a77efd1a3f8549f7f5fdad4ca44
     environment:
       TERM: dumb
       # Limit JVM heap size to prevent exceeding container memory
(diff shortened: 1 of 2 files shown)


MatthewTighe:
That gif is great

corbinlc:
That is how my face looked when someone told me they started block `chdir`.  WTF people !$!

corbinlc:
@MatthewTighe have you seen build_debug fail like this before?

MatthewTighe:
No, that's a new one. Seems like a random failure to me.

.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ reference:
   android_config: &android_config
     working_directory: *workspace
     docker:
-      - image: circleci/android:api-28-ndk
+      - image: circleci/android@sha256:fa7a00c75f4b28cc4f2a15a382fb76e830d62a77efd1a3f8549f7f5fdad4ca44
```
MatthewTighe: This doesn't seem like a permanent solution to me. Maybe add a TODO to investigate this more thoroughly later?

corbinlc:
I added a card for this.

PR #1030, opened 2019-08-29
Update proot

corbinlc:
Take another crack at fixing chdir issues, but also take a crack at fixing stat64 issues.
Also, determine which arch was actually unpacked for the libraries

Just trying to work through some current users issues

UlaFiles.kt

Yes, I ran in debug and made sure it listed out the expected arch on a couple of devices.

They will be able to actually use our app if they are affected by these issues,

I miss you guys


app/src/main/java/tech/ula/model/repositories/AssetRepository.kt
```diff
@@ -24,7 +24,8 @@ data class DownloadMetadata(
 class AssetRepository(
     private val applicationFilesDirPath: String,
     private val assetPreferences: AssetPreferences,
-    private val githubApiClient: GithubApiClient = GithubApiClient(),
+    private val ulaFiles: UlaFiles,
+    private val githubApiClient: GithubApiClient = GithubApiClient(ulaFiles),
```
MatthewTighe: I think the initialization of `GithubApiClient` should probably be handled by whatever is constructing the `AssetRepository`, in order to match code style in the rest of the app. That is to say, we shouldn't use a default parameter here.

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
```diff
@@ -233,7 +233,8 @@ class FilesystemEditFragment : Fragment() {
             navController.popBackStack()
         } else {
             try {
-                filesystem.archType = DeviceArchitecture().getArchType()
+                val ulaFiles = UlaFiles(activityContext, activityContext.applicationInfo.nativeLibraryDir)
```
MatthewTighe: It doesn't look like this needs to be surrounded in a `try catch` block anymore.

corbinlc:
Hey @MatthewTighe, report_coverage is failing every time.  Is this something I broke?

app/src/main/java/tech/ula/ui/FilesystemEditFragment.kt
corbinlc: I removed the `try/catch`

app/src/main/java/tech/ula/model/repositories/AssetRepository.kt
corbinlc: True.  This has been made to follow the style.

MatthewTighe:
No, the coverage reporting issue doesn't look like anything you did. I'd have to look into fixing it.

PR #1075, opened 2019-10-22
New Debian and Kali images

corbinlc:
This version only rolls the version number so people will see if the broken VNC startup will work again.  This was specific to people with devices using the armhf Debian or Kali images and caused the VNC server to never start.  It would just say starting service.  The actually changes are to those two asset sets and not here.


syrranskuk:
Yo

PR #1178, opened 2020-02-07
Added Simplified Chinese translation

xbmanh:
**Note**


anticybercrimer:
Very good

PR #1270, opened 2020-07-30
prompt for donations

corbinlc:
Very basic flow to prompt users for a donation and then prompt them for how much and how often they would like to contribute.  Meant to be very non-intrusive and not ask too frequently.


licaon-kter:
These are donations via Google?

PR #1364, opened 2021-03-10
German translation

pp-sepag:
German translation based on RudiAusBuddeln@043bc1a
Modified some terms to sound more natural.


RudiAusBuddeln:
The problem is that the imperative in app designs sounds much more natural in english. "Delete all sessions and try again" would be "Lösch alle Sitzungen und versuch es erneut", which sounds weird because there is no person and the informal construction is used. The formal one don't sound quite right either in german. Now there is weird mix. In hindsight, I should probably have used the imperative everywhere.

PR #1397, opened 2021-05-29
Ported ctrl-space workaround from latest termux.

yas99en:
**Note**

*  Currentry, UserLAnd terminal does not accept ctrl-space after android9. But the latest termux solved this problem. This PR ports the fix.
* specification
  * In /host-rootfs/data/data/tech.ula/files/home/.termux/termux.properteis, if the property "ctrl-space-workaround" is set true, the workaround is enabled.

I want to use emacs on UserLAnd terminal.

TerminalView.java
It contains the core of this fix.

Yes, It works on my android 11 phone(pixel4a 5g).

End users can use ctrl-space key bind on UserLAnd terminal.
Especially, emacs users needs ctrl-space.

### DIFF
--- a/termux-app/terminal-term/src/main/java/com/termux/app/TermuxPreferences.java
+++ b/termux-app/terminal-term/src/main/java/com/termux/app/TermuxPreferences.java
@@ -47,6 +47,7 @@ final class TermuxPreferences {
     int mBellBehaviour = BELL_VIBRATE;
 
     boolean mBackIsEscape;
+    boolean mUseCtrlSpaceWorkaround;
     boolean mShowExtraKeys;
     
     /**
@@ -57,10 +58,10 @@ final class TermuxPreferences {
     }
 
     TermuxPreferences(Context context) {
+        home_path = context.getFilesDir().getAbsolutePath() + "/home";
+
         reloadFromProperties(context);
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
-
-        home_path = context.getFilesDir().getAbsolutePath() + "/home";
 
         float dipInPixels = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, 1, context.getResources().getDisplayMetrics());
 
@@ -171,6 +172,7 @@ final class TermuxPreferences {
         }
 
         mBackIsEscape = "escape".equals(props.getProperty("back-key", "back"));
+        mUseCtrlSpaceWorkaround = Boolean.parseBoolean(props.getProperty("ctrl-space-workaround"));
 
         shortcuts.clear();
         parseAction("shortcut.create-session", SHORTCUT_ACTION_CREATE_SESSION, props);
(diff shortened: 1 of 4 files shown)


termux-app/terminal-term/src/main/java/com/termux/app/TermuxPreferences.java
```diff
@@ -57,11 +58,11 @@ static int clamp(int value, int min, int max) {
     }
 
     TermuxPreferences(Context context) {
+        home_path = context.getFilesDir().getAbsolutePath() + "/home";
```
yas99en, line 61: reloadFromProperties() requires home_path.

chamaken:
+1 I have the same problem and want to use emacs on UserLAnd terminal too.
see - https://github.com/termux/termux-app/issues/1333

eastsideap3:
> oem flash unlock``

termux-app/terminal-term/src/main/java/com/termux/app/TermuxPreferences.java
```diff
@@ -47,6 +47,7 @@
     int mBellBehaviour = BELL_VIBRATE;
 
     boolean mBackIsEscape;
+    boolean mUseCtrlSpaceWorkaround;
```
eastsideap3, line 50: termux-app/terminal-term/src/main/java/com/termux/app/TermuxPreferences.java

eastsideap3 (approved):

sonwh98:
This problem is still happening in release version 2.8.3 . It makes emacs un-usable

yas99en:
This patch works well on my android phone.
my environment: huawei mate 9, android9, userland 2.8.3

Did you create /host-rootfs/data/data/tech.ula/files/home/.termux/termux.properteis ?

sonwh98:
thanks for responding. i'm using this in vnc lxde desktop. if i use ssh version it works. perhaps this is because the ssh version uses termux?  it doesn't work using lxde desktop via vnc.

sonwh98:
android is intercepting the ctrl-space keys and not sending. since i only use this key shortcut for emacs, i just found out that ctrl-@ also maps setting a mark in emacs. this solution will suffice for now. I hope this gets fixed in android

yas99en:
Hi @sonwh98 , I see you want to use VNC.
First, this patch is for ssh terminal, not for vnc.
If you want to use VNC, you should select the appropriate VNC client.
"VNC Viewer" of Real VNC supports ctrl-space, and bVNC does not support it.

sonwh98:
@yas99en wow awesome! I just confirmed that real VNC client does send ctrl-space . This is awesome now have a complete Linux workstation in my pocket. Only thing missing to replace my laptop is this https://www.aliexpress.com/item/1005001380819089.html?gatewayAdapt=glo2vnm&fbclid=IwAR1aw_KaG0i80h7TVBffq4Zno9PqPCrHTrWrGVOPnasLPrgCqpV4a3Xc0N8

